PR #2, opened 2010-09-10
mkdir: cannot create directory `/var/run/riak/': Permission denied

retr0h:
Need to sudo before switching users

Ran into the same bug reported here:
  http://pastebin.com/DR85ZDCy

Updated handler script to sudo mkdir && sudo chown before switching
over to $RUNNER_USER.

John


slfritchie:
John, I noticed that your branch is based on code from 19 Aug 2010.  There were some packaging changes for the 0.13 release that might've addressed this problem in a different way.  Would you mind re-checking with a newer package and, if it's still broken, then submit a new pull request?  (We really do want to fix bugs like this, if they still exist in the latest release.) We've had too many problems with Debian vs. Red Hat vs. Fedora vs. Solaris packaging bits to try patching older code without a Master's thesis-worth of code study to prepare for it.....

-Scott

PR #4, opened 2010-10-28
Fix for the 'distdir' build target in the Makefile

OJ:
Hi guys,

I noticed that the Makefile still had a few hangovers from when mercurial was being used. First of all, if you cloned the repo and ran 'make' you'd this error:

abort: There is no Mercurial repository here (.hg not found)

Which was the result of RIAK_TAG using hg to identify what the current tag is. Also, the 'distdir' target itself had lots of mercurial-related stuff. I took a few guesses based on what was there already and attempted to implement the same functionality using git. I've also added some code which is able to "archive" both git and hg repos, which is handy as dependencies might live in mercurial repositories anyway.

Let me know what you think. Feedback is greatly appreciated.

All the best!
OJ


dizzyd:
OJ, thanks for the feedback. We're going to be doing some work on the make system as we expose the mechanisms for building debs/rpms. I'll take a look at the specifics of your patch next week (Nov 8th) -- I'm traveling out to SFO right now.

Thanks!

OJ:
Hey Dizzy,

Thanks for the response mate. I had a feeling that you would look to use something more complete down the track, this was really an attempt at helping to fill the gap in the mean time until you guys had something more solid in place. Certainly no pressure to accept it. Only if you find it useful! (at least it gets rid of the errors you see when making, not that it was hard to do that :)).

Have a safe trip mate.
Cheers

OJ

OJ:
Hey Dizzy, did you get a chance to review this? Is it something that can at least help in the interim?

Cheers dude.

rtilder:
Hey, OJ.  I wound up doing a partial pull of your changes limited to the Makefile targets.  I changed the manner in which the RIAK_TAG is defined in order to work with versions of git prior to v1.6.2(when the git tags --contains flags appeared).  Many thanks.

--Ryan

OJ:
Hey Ryan,

Thanks for the feedback mate. I totally understand your change. I speculated a lot on what you would have wanted and would have been surprised if I had managed to produce something that made it to the mainline without modification. I am happy if I've managed to help in any way.

Thanks for the response, all the best!
OJ

PR #5, opened 2010-11-02
Riak documentaion elaboration: riak_kv_lru, vnode_cache_entries

jlouis:
You may want to pull this documentation elaboration to riak. A reason for not doing it could be to hide the knob from the user - so they don't go twiddling with knobs they should not play with.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -67,6 +67,13 @@
             %% Value is in milliseconds and defaults to 1000ms if not set.
             %{vnode_mr_timeout, 1000},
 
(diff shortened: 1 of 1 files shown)


slfritchie:
Jesper, I've been informed by informed sources that that config knob will be disappearing in the 0.14 release, due out in a few weeks.  Thanks for the docs (I was about to add your text to https://wiki.basho.com/display/RIAK/Configuration+Files also), but it looks like this batch won't be necessary.

-Scott

PR #10, opened 2010-12-02
Bashism

ghost:


slfritchie:
Tuncer, I'm a big fan of removing bashisms, thanks.  I'll duplicate these in our other major packages also.

-Scott

PR #12, opened 2010-12-06
Fix whitespace error and add file local variables

ghost:


dizzyd:
Merged into riak_kv-0.14.

PR #16, opened 2010-12-14
dss include gitvsn (take 2)

dizzyd:
Include git.vsn when constructing package archives. This makes it possible for us to know what version of a given sub-repo was used during compilation and not have to do so many RCs.


rtilder:
+1

I bite my thumb at Scott's FreeBSD leanings with respect to $(call ...) semantics but  his second comment is something we should bug or later correction.

PR #19, opened 2011-01-28
Bz971 stats latency

slfritchie:
Pull request 1 of 3 for bz971.  Rebar.config comments will be yanked out before merge to master.


jonmeredith:
Should the stats dep actually be against riak_core instead?  That way we wouldn't need to adjust deps for riak, riak_ee and riak_search

jonmeredith:
Still need to remove the comments above - don't forget ee & search :)

PR #1, opened 2010-09-03
Way improved after a bit of sleep.  Ignore previous pull request.

b:

PR #20, opened 2011-02-03
Bz997 erlcrashdump

seancribbs:
Puts the erl_crash.dump file in a writeable place for each platform so we don't lose information about hard crashes.

### DIFF
--- a/package/deb/vars.config
+++ b/package/deb/vars.config
@@ -10,6 +10,7 @@
 {sasl_log_dir, "/var/log/riak/sasl"}.
 % vm.args
 {node,         "riak@127.0.0.1"}.
+{crash_dump,   "/var/log/riak/erl_crash.dump"}.
 % bin/riak*
 {runner_script_dir,  "/usr/sbin"}.
 {runner_base_dir,    "/usr/lib/riak"}.
--- a/package/rpm/SPECS/riak.spec
+++ b/package/rpm/SPECS/riak.spec
@@ -44,7 +44,7 @@ cat > rel/vars.config <<EOF
 {hook_js_vms, 2}.
 % vm.args
 {node,         "riak@127.0.0.1"}.
-{crash_dump,   "/var/log/riak/erl_crash.dump"}.
+{crash_dump,   "%{_localstatedir}/log/%{name}/erl_crash.dump"}.
 % bin/riak*
 {runner_script_dir,  "/usr/sbin"}.
 {runner_base_dir,    "%{riak_lib}"}.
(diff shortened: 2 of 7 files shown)


jonmeredith:
+1 - need to do for ee/search too.

slfritchie:
Sean: I haven't been able to figure out what & when does the
{{whatever_variable}} text substitution for Solaris & OpenSolaris
packages ... but the same thing should be done for them.  If
you had a solution for *Solaris, I'd merge it pronto.  :-)

A version of the 0f199e4 patch for the Debian and *Solaris would
also be useful: it'd be nice to avoid contributing to platform-specific
drift.

One last thing ... a static file name runs the risk of the following
sequence of events (which has bitten me in the past):
1. Bad crash happens, file "foodump" created.
2. Customer doesn't react fast enough, another unrelated crash
    happens and writes a new "foodump".
3. Customer calls for support.
4. "foodump" doesn't contain anything related to step #1's crash.

seancribbs:
The alternative to what you mention is to have it be dumped in the directory where the emulator is running, which is what caused lost crash dumps in first place (/usr/lib/riak is not writeable).  From the erts manual (http://erlang.org/doc/apps/erts/crash_dump.html):

"The system will write the crash dump in the current directory of the emulator or in the file pointed out by the environment variable (whatever that means on the current operating system) ERL_CRASH_DUMP. For a crash dump to be written, there has to be a writable file system mounted."

slfritchie:
Yup, but we don't have all platforms using the same working dir
either.  Bah.  I'd love to change things to that the VM's working
dir was /path/to/riak/data dir, but I haven't the foggiest notion if
that'd break an existing installation or not.

seancribbs:
Scott: is there a way to tell the emulator from vm.args or app.config to chdir after starting?

slfritchie:
No, not that I'm aware of.  We could change it, but we also don't have
a config knob that says where that dir is.  IIRC the best that's currently
there is to take bitcask's dir and move up to its parent.  Except for
those configs that don't use bitcask.  So we'd have to add it.  Which I
think would be good, but hasn't been done yet.  </teeth mode=gnashing>

seancribbs:
Surely we're not the first to run into this. Maybe a patch to OTP is in order as well?

slfritchie:
Well ... from the OTP team's point of view, the cwd isn't a sharable
thing.  If you have two OTP apps that want different cwds, one of them
(at least) is going to be sad.

seancribbs:
This has already been applied elsewhere.

PR #24, opened 2011-03-21
Make 'riak ping' exit with non-zero status if node doesn't respond

chids:
I suggest making "riak ping" exit with non-zero status if the node doesn't respond to ease conditional scripting based on node status etc.


jonmeredith:
Thanks for the patch.  We're refactoring some of the riak scripts at the moment and a new version that returns non-zero for lots of failures should be available soon.

PR #26, opened 2011-03-29
Bz710 riak init script exit status

Vagabond:


jonmeredith:
+1 to merge.

Checked start / stop / restart / reboot / attach all work.
Ran up another node on the same handoff ports and start fails now with non-zero exit rather than the node crashing silently.

jonmeredith:
abort, abort.
Wait until the packaging refactor is complete.

PR #28, opened 2011-04-21
Bz1077 nodetool ssl support

slfritchie:
Basho reviewer: dev or DA, someone familiar with Riak packaging is a plus.

The new OTP app env vars to riak_core, named platform_*_dir, aren't used
directly by nodetool, but they're things that I've long wanted basho_expect to
have access to.  Since I was mucking around with the rebar template stuff, I
added them.

Nodetool needs to use the same SSL config items (from the "vm.args" file)
that the riak VM needs, except that it can't use the "vm.args" file as-is because
it contains stuff that nodetool cannot use, e.g. "-name foo@bar".  Therefore,
I use a "sed" hack to strip out the SSL config stuff from "vm.args" and put it
into a temp file that nodetool can use.  Each time the "riak" script is run, that
temp file is created, which should allow "vm.args" to be the single place
where this config info should be maintained.

If this packaging looks sane, then I'll apply the same changes to RiakEE and
RiakSearch packaging.  I haven't done so yet because reviewers might have
substantial recommentations for more work.

Possible todo items:

```
1. Put the docs in the 2nd commit into another, more accessible place
1. Is there a better style/method for managing the key files?
```

### DIFF
--- a/package/deb/vars.config
+++ b/package/deb/vars.config
@@ -8,6 +8,14 @@
 {bitcask_data_root, "/var/lib/riak/bitcask"}.
 {sasl_error_log, "/var/log/riak/sasl-error.log"}.
 {sasl_log_dir, "/var/log/riak/sasl"}.
+
+% Platform-specific installation paths
+{platform_bin_dir, "/usr/sbin"}.
+{platform_data_dir, "/var/lib/riak"}.
+{platform_etc_dir, "/etc/riak"}.
+{platform_lib_dir, "/usr/lib/riak"}.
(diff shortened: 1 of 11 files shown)


dizzyd:
This is overlapping with some work I've been doing w/ node_package and our private repos. Let's touch base in the morning and discuss.

seancribbs:
Documentation? SSL is always persnickety about what it allows, so knowing the purpose and format of each certificate would be helpful (also how to generate them if necessary).

dizzyd:
+1. I'll look into incorporating these ideas into node_package as well.

PR #29, opened 2011-04-21
bz1075 Add riak_kv param to set backlog option on pb listen socket

russelldb:
When the java pb client creates a bunch of simultaneous connections to riak pb interface a few of those connections get dropped by riak.

Upping the backlog mitigates this. 
Add documentation and example of pb_backlog env parameter to app.config
See also https://github.com/basho/riak_kv/pull/81

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -40,6 +40,13 @@
             %% will bind to
             {pb_port, {{pb_port}} },
 
+            %% pb_backlog is the maximum length to which the queue of pending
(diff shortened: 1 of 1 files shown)


seancribbs:
@russelldb When PB refactor lands, this will be obselete. OTOH I think basho/riak_kv#81 has already landed?

russelldb:
yeah, is old, and dead, closing now, thanks.

PR #31, opened 2011-05-17
make four dev nodes, not three

beerriot:
four nodes produce slightly different cluster behavior, since they
divide a power-of-two partition count evenly; useful for testing


dizzyd:
+1

PR #35, opened 2011-06-06
Az452 integrate search with riak

rzezeski:
Moving forward Search will come bundled w/ Riak as a feature that can be enabled (not by default).

See also https://github.com/basho/riak_search/pull/63

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -123,17 +123,6 @@
                 {enabled, false}
                ]},
 
- %% Qilr Config
- {qilr, [
-         %% NOTE: Change with 0.14 release: by default, JVM text analyzer is
-         %% disabled via missing/commented 'analysis_port' option or
-         %% via {analysis_port, undefined}.
-         %% Uncomment the following line in order to use JVM Lucene-based text analyzers.
-         %% {analysis_port, {{analyzer_port}} },
-
-         {backend_conn_count, {{analyzer_conns}} }
-        ]},
-
  %% Merge Index Config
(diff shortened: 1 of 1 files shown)


rel/vars/dev2_vars.config
```diff
@@ -15,6 +15,11 @@
 {sasl_log_dir,      "log/sasl"}.
 {mapred_queue_dir, "data/mr_queue"}.
 
+%% riak_search
+{merge_index_data_root,  "data/merge_index"}.
+{analyzer_conns,    50}.
+{analyzer_port,     6095}.
```
Vagabond, line 21: Shouldn't the analyzer port be different across the 3 different dev VM configs so they don't conflict?

tests/riak_search/_files/books/books.xml
```diff
@@ -0,0 +1,47 @@
+<add>
```
Vagabond, line 1: Why aren't all these test files bundled with riak_search?

rel/vars/dev2_vars.config
rzezeski, line 21: They should, but considering this goes away w/ qilr I'll just remove it in the Qilr patch

tests/riak_search/_files/books/books.xml
rzezeski, line 1: Because the only way to run them is with an instance of Search running, which is now part of Riak.

Vagabond:
+1 to merge

PR #37, opened 2011-06-15
Add merge_index to reltool.config

dreverri:
Fixes:
==> rel (generate)
ERROR: Unable to generate spec: Undefined applications: [merge_index]

make: **\* [rel] Error 1


rzezeski:
It's been a long time since I did much work with releases but AFICT this PR isn't needed.
1. Jared woud yell if the build was failing in this way
2. It seems reltool automatically pulls in applications that listed apps depend on, and since riak_search is listed it will pull in merge_index

Any objects on closing Dan?

dreverri:
I am no longer able to reproduce this issue.

PR #39, opened 2011-06-23
Better comment for cookie value

rzezeski:
Recently a user was confused about the purpose of the cookie value and
set it to a unique value for each node.  I thought clearing this up in
the inline comment might be useful.


jtuple:
+1

jaredmorrow:
+1

PR #40, opened 2011-06-28
Add base config for eleveldb driver

dizzyd:
Depends on https://github.com/basho/riak_kv/pull/123

### DIFF
--- a/package/deb/vars.config
+++ b/package/deb/vars.config
@@ -6,6 +6,7 @@
 {pb_ip,        "127.0.0.1"}.
 {pb_port,      8087}.
 {bitcask_data_root, "/var/lib/riak/bitcask"}.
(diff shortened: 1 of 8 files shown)


slfritchie:
After changing `etc/app.config` for `rel/riak` and `dev/dev?` etc dirs,
all of the changes work, for `make stage` and 'make stagedevrel`and
then running a single node (former) and 4 nodes joined together into
a single cluster (latter), then storing 10K keys, fetching them, and then
looking in the appropriate`.../data/leveldb` subdirectory to verify that
leveldb's files are actually there.  +1 to merge, thanks!

dizzyd:
Thanks, Scott. Will merge once I've worked out some compilation issues in leveldb.

PR #41, opened 2011-07-01
Sanity check the app.config file before starting riak

Vagabond:
Run file:consult on the app.config before trying to start riak. This prevents the issue where 'riak start' silently fails if the config file is invalid in some way.


slfritchie:
Comments from IM:
- Go whole hog with Apache httpd-style 'chkconfig'
- Do the same for RiakEE's packaging.

Vagabond:
I hosed up this pull request, reopening

PR #43, opened 2011-07-18
Az545 search update rpm

rzezeski:
Update the RPM package now that Search is integrated into Riak.


jaredmorrow:
cherry-picked master's rebar so I can test the builds

jaredmorrow:
+1, confirmed the package contents and ownership

PR #46, opened 2011-07-20
Az550 chmod bindir deb

jaredmorrow:
I added 0755 perms to both files/erl and files/nodetool so outside of any package manager they'd have the correct permissions.   In addition, the debian packaging didn't ensure the binaries were executable, so that was added.

I tried several different ways, so that's why there are a few commits.


beerriot:
+1: fixed the 'unable to open nodetool' error I was seeing

PR #49, opened 2011-07-27
set mapred_system=pipe by default in new app.config

beerriot:
Now that riak_pipe is included in master, we could switch to using it by default.  This patch does so, and also comments out the legacy-specific MapReduce configs in app.config as well.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -68,9 +68,16 @@
             %% mapred_name is URL used to submit map/reduce requests to Riak.
             {mapred_name, "mapred"},
 
+            %% mapred_system indicates which version of the MapReduce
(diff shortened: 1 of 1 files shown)


slfritchie:
> Are we ready to have pipe servicing MapReduce requests by default?

Yes, I think we should flip the switch ... but I think if you could wait until
merging the branch until Wednesday late afternoon or so ... there are
some packaging details and basho_expect testing that I'd like to do with
basho_expect first, before this config change.

Otherwise, ja, let's do it.  Could you also make this change to the RiakEE
repo?

beerriot:
You got it.  I'll do them both either last thing on Wednesday or first thing on Tuesday.

PR #50, opened 2011-07-28
riak_pipe build dependency for rebar.

ncb000gt:
After trying to `make rel` and getting an error I noticed that riak_pipe wasn't being pulled in. I added it to `rebar.config`. I'm assuming this was the place for it and not a sub dependency.


ncb000gt:
I built this on a different machine without my change and noticed riak_pipe already being pulled in by the appropriate sub dependencies. I guess my other machine was out of sync or something else was going on. Either way I don't think this is needed.

PR #52, opened 2011-08-01
AZ567 - Update 2i merge_index data_root setting.

rustyio:
With 2i, we now run an additional merge_index instances. This change sets the 2i instances to point to a different data_root than the search instances.


jonmeredith:
+1 merge.  There may be a clash with the default backend on merge - please make sure bitcask is the default backend for now.

PR #55, opened 2011-08-04
added 5th dev node

jkirkell:
I added a 5th node to the makefile and an associated dev5 config file. It seems like a better number.


Vagabond:
I did this last week too, except I needed 6 nodes. Anything over N+1 is definitely handy.

jkirkell:
thanks

PR #57, opened 2011-08-08
Added sudo-dependency for Debian packages

msiebuhr:
The command-line utilities `riak` and `riak-admin` uses `sudo`, so it should be a dependency...

### DIFF
--- a/package/deb/control
+++ b/package/deb/control
@@ -8,7 +8,7 @@ Homepage: http://basho.com/
 
 Package: riak
 Architecture: any
-Depends: adduser, logrotate, ${shlibs:Depends}, ${misc:Depends}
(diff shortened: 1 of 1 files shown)


PharkMillups:
Thanks, Morten! We'll check it out when we have a moment. 

Mark

msiebuhr:
BTW: This should fix [bug 1130](https://issues.basho.com/show_bug.cgi?id=1130).

Related: I've been looking for Riak-Search's ditto file (it also misses the `sudo`-dependency), but I can't seem to find it.

msiebuhr:
Hi Mark & al,

Just a quick reminder on this (fairly modest) patch, which I guess has been lost on the recent 1.0-work.

Just for good measure; the patch applies cleanly against the riak (currently 16e0f31a203491ef1b8084bffbece92fa724a7d7).

PR #58, opened 2011-08-08
Add Search test cases to verify deletion of proxy objects

rzezeski:
bz1147

### DIFF
--- a/tests/riak_search/_files/docs/doc1
+++ b/tests/riak_search/_files/docs/doc1
@@ -1,0 +1,1 @@
+The quick brown fox jumps over the lazy dog.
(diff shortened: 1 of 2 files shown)


jtuple:
The first exists check in "Verify proxy object is deleted for local docs" fails for me. All the other exists checks work fine.

...
(PASS) {echo,   "Verify proxy object is deleted for local docs"},
(PASS) {index,  "../_files/docs"},
(-FAIL-) {exists, <<"_rsid_test">>, <<"doc1">>, true},
(PASS) {delete, "../_files/docs"},
(PASS) {exists, <<"_rsid_test">>, <<"doc1">>, false},
...

rzezeski:
Hrm, worked for me just now, ran it 5 times.  I wonder if it's a timing thing?  Could you try putting an echo or two between the delete and exists check and see if that makes a difference.

jtuple:
All good, merge away.

PR #60, opened 2011-08-11
Az520 md5 checksum on package

jaredmorrow:
We are getting upload errors to our build hosting on S3 so this creates md5 sums for each package we build.


seancribbs:
Dare I ask for SHA256 hashes as well? The Chef cookbook uses them.

jaredmorrow:
Oh, I can look into that.   I just checked on solaris and digest can support sha256.   If no one minds, I'll just switch md5->sha256 so we aren't creating too many files.   Opinions?

jaredmorrow:
Changed to sha256 and changed deb/makefile to be more like rpm/makefile

dizzyd:
+1

PR #65, opened 2011-08-31
Duplicating sean's efforts in commit 1abc429c16c807f7c011 (BZ997) for ria

massung:
Duplicating sean's efforts in commit 1abc429c16c807f7c011 (BZ997) for riak/master and added crash_dump variable for dev4_vars.config as well.

### DIFF
--- a/package/deb/vars.config
+++ b/package/deb/vars.config
@@ -27,6 +27,7 @@
 
 % vm.args
 {node,         "riak@127.0.0.1"}.
+{crash_dump,   "{{platform_log_dir}}/erl_crash.dump"}.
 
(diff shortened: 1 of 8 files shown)


slfritchie:
Hrm, looks like I took care of applying this patch to riak_ee "master"
while on a packaging-unification rampage back in July.

+1 to merge, thanks.

Bug 997 also mentions the 0.14 branches of Riak and RiakEE.  Separate
pull requests for each of those branches isn't required, but please apply
the patches to those branches and note the commits in the bug ticket.
0.14 may live a bit longer, who knows, so might as well apply it as Dizzy
requests.

PR #68, opened 2011-09-06
Change default of legacy_keylisting option in app.config to false.

kellymclaughlin:


rel/files/app.config
```diff
@@ -132,11 +132,8 @@
             {vnode_vclocks, true},
 
             %% This option enables compatability of bucket and key listing
-            %% with 0.14 and earlier versions. Once a rolling upgrade to
-            %% a version > 0.14 is completed for a cluster, this should be 
-            %% set to false for improved performance for bucket and key
-            %% listing operations.
-            {legacy_keylisting, true}
+            %% with 0.14 and earlier versions.
```
Vagabond: I think you removed a little too much from the comment here. The note saying that after the rolling upgrade is complete you should change this to false is useful information.

kellymclaughlin: I thought maybe I was being too verbose. I'll add it back.

Vagabond:
+1 on this.

Vagabond:
This was merged but not closed on GH.

PR #73, opened 2011-09-15
Add a backend_status command to riak-admin.

kellymclaughlin:
Add a backend_status command to riak-admin that fetches the status of
the backend for each available vnode. The vnode index, the backend,
and the data returned from calling the backend module's status
function are displayed.


kellymclaughlin:
Changed the `backend_status` command to the more general `vnode-status`. `vnode-status` retrieves the local vnodes instead of using coverage to get the status of the cluster vnodes.

jonmeredith:
+1 merge.

PR #76, opened 2011-09-21
Fix bz1195; output the most recent tag to vsn.git for rebar to use

dizzyd:


jaredmorrow:
+1, git.vsn files now populated, merging for 1.0 rc1

PR #80, opened 2011-09-27
Create platform_data_dir in the riak startup script

jaredmorrow:
Before accessing the platform_data_dir in the riak startup script, ensure it exists.

Fixes: bz1224


dizzyd:
+1

PR #83, opened 2011-09-29
Change 'riak-admin remove' to 'riak-admin force-remove'

jtuple:
riak-admin remove in 1.0 immediately drops replicas, which is different behavior than remove in 0.14.2. To ensure users are aware of this change, remove is renamed to force-remove and a message is printed if the old remove command is used (directing user to either leave or force-remove).


Vagabond:
+1 to merge

PR #90, opened 2011-10-05
BSD support

Vagabond:
Don't use /bin/bash in the shebang, use /bin/sh (and add a crazy workaround for Solaris' /bin/sh not actually being a POSIX shell).

### DIFF
--- a/rel/files/erl
+++ b/rel/files/erl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 ## This script replaces the default "erl" in erts-VSN/bin. This is necessary
 ## as escript depends on erl and in turn, erl depends on having access to a
--- a/rel/files/erl
+++ b/rel/files/erl
@@ -1,5 +1,14 @@
 #!/bin/sh
 
+# /bin/sh on Solaris is not a POSIX compatible shell, but /usr/bin/ksh is.
+if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
+    POSIX_SHELL="true"
+    export POSIX_SHELL
+    exec /usr/bin/ksh $0 $@
+fi
+unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
+
+
 ## This script replaces the default "erl" in erts-VSN/bin. This is necessary
 ## as escript depends on erl and in turn, erl depends on having access to a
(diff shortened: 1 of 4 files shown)


dizzyd:
+1, if it passes b_e

zaa:
Works ok for us on 

```
uname -mrs
FreeBSD 8.1-RELEASE i386
```

jj1bdx:
At least about the shebang script, it looks OK to me.

(I changed it to /usr/local/bin/bash when I tested, but this is not recommended because for FreeBSD bash is only a Port and not with the OS distribution.)

sc68cal:
What about using `env` to locate bash - that is typically how to ensure portability.

jj1bdx:
For FreeBSD and other BSD-derived OSes, bash is _not_ installed as a part of the distributions. So shebang of `#!/usr/bin/env bash` does not work anyway.
If there were anything critical in the script needed something only bash could do, the usage of bash would be justified, but I think that's not the case for this pull-request.

sc68cal:
True, but the `Makefile` for the port can just add

```
RUN_DEPENDS=    bash:${PORTSDIR}/shells/bash
```

jj1bdx:
I think adding a bash dependency to the FreeBSD Port Makefile does not solve the fundamental issues of bash dependencies,  though it is surely a workaround for the time being. The question is whether we have to require Riak to install bash or not. And not all people want to install Riak from the source code.

FYI, on FreeBSD Port for Erlang (`lang/erlang`), bash is not a requirement. And I think it shouldn't be.

Vagabond:
The bash dependency is pointless and should/will be removed, I just haven't had the time to reverify that the change doesn't break anything.

koobs:
Confirming a successful build and test using the following port Makefile and patch: https://gist.github.com/ee3563208933cd8e97a6 

Note the post-build: target, which blanket replaces /bin/bash references with /bin/sh refs with no other modifications.

./riak console output: http://codepad.org/Co9W7spF

All ./riak and ./riak-admin commands seem to work.

The patch in the gist above is for code already committed but didn't make it into 1.1.2 so shouldn't be required in the port for long

Port is based on draco2002 original work at https://github.com/draco2003/freebsd-database-riak

Only work left is to add a pkg_plist, install /rel/\* bits to appropriate system locations and get an RC script in.

jaredmorrow:
+1, tested that everything still works on Solaris, Fedora, Ubuntu

PR #91, opened 2011-10-05
remove dusty mapred docs, in favor of wiki.basho.com

beerriot:
https://issues.basho.com/15 requested that these two docs be merged into one ages ago.  I propose removing them now, in favor of http://wiki.basho.com/MapReduce.html


dizzyd:
+1, if we have a reference to the wiki somewhere in the docs shipped with source...

beerriot:
Good point, @dizzyd.  I've added a link in README.org.  I also propose removing README, since GitHub now renders .org, and README is lagging anyway.

dizzyd:
+1

PR #93, opened 2011-10-16
Link to the bug tracker in README. 

timoxley:
Link to the bug tracker in README. Is disconcerting to find github project with no issue queue and no link to substitute.


jonmeredith:
+1 merge - thank you.

PR #95, opened 2011-10-19
Use nodetool's rpc_infinity method when calling cluster_info

slfritchie:
Subject says it all, almost: without this patch, a 60 second timeout is
used, which is far too short for large clusters and/or slow nodes.


dizzyd:
+1

PR #98, opened 2011-10-28
Add 'riak version' command

Vagabond:
At package build time, capture info about when the package was built, what version it is, when it was built and what OS it was built for. Then expose that information in a machine and human parsable format via 'riak version'


jaredmorrow:
+1

PR #99, opened 2011-10-31
Initial changes for riak_control.

massung:


rel/files/cert.pem
```diff
@@ -0,0 +1,17 @@
+-----BEGIN CERTIFICATE-----
```
Vagabond, line 1: Committing certificates like this is probably not a good idea. At the very least they should be generated, no?

massung, line 1: Nope, can't create them in the Makefile since the user has to interact with the program to actually generate them. Dizzy and Jon both seemed fine w/ checking in a self-signed certificate. I'm conflicted. On one hand you want this to be dead simple for someone to launch and get running. On the other hand, you have all the fun security stuff. 

One idea I had - but am unsure if it would work - was to have the makefile create soft links to the OS's default certs, or perhaps to some created by mochiweb? But I'm not sure if that's a bad idea or not.

Vagabond, line 1: You can automate creating certs by using a script that feeds data into the interactive command. I've done it before. I bet an escript that spawns a port would work fine. There even seems to be an example of doing this shipped with erlang.

See also:
http://www.erlang.org/documentation/doc-5.7.5/lib/ssl-3.10.8/doc/html/create_certs.html
https://github.com/erlang/otp/blob/master/lib/ssl/examples/certs/src/make_certs.erl
https://github.com/erlang/otp/blob/master/lib/ssl/test/erl_make_certs.erl

jonmeredith:
+1 merge - agree the certs should probably be auto-generated, we can file it as an additional task to convert.  It doesn't need to hold up the merge.

PR #100, opened 2011-11-03
Fix init script (/var/run/riak, status, start)

dhull:
This pull request fixes three small problems with riak's RPM init script:
1. On Fedora 15, /var/run is a tmpfs filesystem, and so the /var/run/riak directory goes away when the machine is rebooted. Change the init script to create it if it doesn't exist. This corresponds to bug 775 (https://issues.basho.com/show_bug.cgi?id=775) in the basho bugzilla.
2. Add a "status" command.
3. The "start" command incorrectly detects that riak is not running even if it is.

### DIFF
--- a/package/rpm/SOURCES/riak_init
+++ b/package/rpm/SOURCES/riak_init
@@ -33,7 +33,7 @@ DAEMON=/usr/sbin/$NAME
 
 start() {
         #Check if Riak is running
-        RETVAL=`$DAEMON ping`
+	RETVAL=`su - riak -c "$DAEMON ping"`
         [ "$RETVAL" = "pong" ] && echo "Riak is already running" && return 1
 
 	# Start daemons.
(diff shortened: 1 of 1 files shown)


package/rpm/SOURCES/riak_init
```diff
@@ -28,7 +33,7 @@ DAEMON=/usr/sbin/$NAME
 
 start() {
         #Check if Riak is running
-        RETVAL=`$DAEMON ping`
+	RETVAL=`su - riak -c "$DAEMON ping"`
```
Vagabond, line 36: Indentation is wrong here.

dhull, line 36: The existing indentation in this file is a hodgepodge of spaces and tabs (sometimes with both on the same line). Previously this line had eight spaces. I replaced them with a tab. It lines up properly if a tab == eight spaces, which is the equivalence that the rest of the file assumes.

jaredmorrow:
WIll merge and fix spacing in a follow on commit.

PR #104, opened 2011-11-21
Generate SSL certs automatically

seancribbs:
Using the instructions from http://www.akadia.com/services/ssh_test_certificate.html, these Makefile targets generate SSL certificates for use in local testing, especially useful for riak_control.


seancribbs:
Closing since the rel templates include a default certificate and key.

PR #107, opened 2011-12-14
Enable new listkeys backpressure for new clusters, by default

beerriot:
If https://github.com/basho/riak_kv/pull/263 is accepted, this patch will enable the new behavior by default.


kellymclaughlin:
+1 to merge

PR #109, opened 2011-12-23
Remove luwak dependency from riak

jaredmorrow:


reiddraper:
`make rel` and `make stage` fail right now with

```
{"init terminating in do_boot","Release riak uses non existing application skerl"}
```

I think it's because `skerl` is still listed in `reltool.config`, but it's no longer available (it was listed as a dep in luwak). Removing them seems to let Riak compile correctly.

reiddraper:
+1

PR #110, opened 2012-01-04
Add "riak-admin top" subcommand

slfritchie:
Usage: riak-admin top [-interval N] \
                      [-sort reductions|memory|msg_q} \
                      [-lines N]

Also, added a new directory to the packaging, lib/basho-patches,
for patched .beam files.  The packaging changes compile and copy
a modified etop_txt.beam file into this new dir.

NOTE: This new lib/basho-patches directory is added to the code
server's path with the "-pa" flag, so it can be used to patch
any non-sticky BEAM file.

Edits to src/etop_txt.erl change the column widths for several
columns to be more useful to humans: since the output is wider
than 80 colums, we make the new output just shy of 132 columns.

### DIFF
--- a/ebin/riak.app
+++ b/ebin/riak.app
@@ -1,0 +1,3 @@
+{application,riak,
+             [{description,"Riak extremely-early boot items"},
+              {modules,[etop_txt]}]}.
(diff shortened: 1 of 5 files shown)


Vagabond:
It'd be really nice if you could do something like `stty size | awk '{ print $1 - 9}'` to auto-size the etop output to the terminal (9 is the number of lines of overhead that are printed, plus the final line).

slfritchie:
Hrm, yes, it could be nice ... but then:
- If someone's window isn't wide enough, what should we do?
- Not all ssh clients seem to pass window size through

Vagabond:
You have my +1, sir.

PR #111, opened 2012-01-04
Az1035 add etop 1.0

slfritchie:
Usage: riak-admin top [-interval N] \
                      [-sort reductions|memory|msg_q} \
                      [-lines N]

Also, added a new directory to the packaging, lib/basho-patches,
for patched .beam files.  The packaging changes compile and copy
a modified etop_txt.beam file into this new dir.

NOTE: This new lib/basho-patches directory is added to the code
server's path with the "-pa" flag, so it can be used to patch
any non-sticky BEAM file.

Edits to src/etop_txt.erl change the column widths for several
columns to be more useful to humans: since the output is wider
than 80 colums, we make the new output just shy of 132 columns.

### DIFF
--- a/ebin/riak.app
+++ b/ebin/riak.app
@@ -1,0 +1,3 @@
+{application,riak,
+             [{description,"Riak extremely-early boot items"},
+              {modules,[etop_txt]}]}.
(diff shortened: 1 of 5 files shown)


Vagabond:
+1

PR #114, opened 2012-01-30
Generate explicit build tags when building off of a non-tagged commit

jaredmorrow:


jonmeredith:
+1 for package/distribution stuff.

Take a look at the changes we discussed to use '|| echo FAIL' on the eunit test case to make sure all tests are run.

Makefile
```diff
@@ -13,14 +10,26 @@ compile:
...
 	./rebar delete-deps
 
-test:
-	./rebar skip_deps=true eunit
+
+TEST_LOG_FILE := eunit.log
+testclean:
+	@rm -f $(TEST_LOG_FILE)
+
+# Test each dependency individually in its own VM
```
jonmeredith, line 24: L31/L30 should be swapped, with `|| echo "Eunit: root FAILED" >> $(TEST_LOG_FILE);)`  after the rebar eunit

jonmeredith, line 24: Otherwise a top level failure will not be spotted and the output is weird

```

==> rel (eunit)
==> riak (eunit)
Eunit: basho_stats FAILED
Eunit: riak_kv FAILED
Eunit: riak_pipe FAILED
./rebar eunit skip_deps=true
==> lucene_parser (eunit)
  All 91 tests passed.
Cover analysis: /Users/jmeredith/basho/work/1.1/riak/deps/riak_search/apps/lucene_parser/.eunit/index.html
==> rel (eunit)
==> riak (eunit)
  There were no tests to run.
Cover analysis: /Users/jmeredith/basho/work/1.1/riak/.eunit/index.html
make: *** [test] Error 3
```

jonmeredith:
One last minor tweak then +1 to merge.

PR #115, opened 2012-02-02
Fix eunit testing so that passing tests don't cause make to fail

jaredmorrow:
When all tests pass, there is no eunit.log and that was causing 'test -s' to exit with return 1.   This just wraps that all in an if clause.


jtuple:
Looks reasonable, +1

PR #117, opened 2012-02-09
vsn.git files are required for all deps when removing .git directories

jaredmorrow:
1.1 RC1 had an issue when trying to build from the release tarball.   Rebar checks for .git directories, and when those are not found it looks for vsn.git files instead to find the tag.   

This file was deleted prior to 1.1 because all tags were collected into a single file.  Now both the single file and each dependency file is created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ build_clean_dir = cd distdir/$(CLONEDIR) && \
                   for dep in deps/*; do \
                       cd $${dep} && \
                       $(call archive_git,$${dep},../../../$(DISTNAME)) && \
-                      mkdir -p ../../../$(DISTNAME)/$${dep}}/priv && \
+                      mkdir -p ../../../$(DISTNAME)/$${dep}/priv && \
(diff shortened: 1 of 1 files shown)


jonmeredith:
+1 merge, don't forget the EEs

drnic:
Do we need to update the copy of rebar for this to work?

The one that is in riak-1.1.4 is from Oct 2010.

```
# ./rebar -V
rebar version: 2 date: 20111027_031525 vcs: git 9197e70

# git show 9197e70     
commit 9197e70bd7f442576eb76fab4011fa36b8739960
Author: Tuncer Ayaz <tuncer.ayaz@gmail.com>
Date:   Wed Oct 26 23:23:07 2011 +0200
```

riak-1.1.4 tar.gz is filled with vsn.git files in the deps, but I think `./rebar get-deps` is still trying to run git commands?

```
make: git: Command not found
./rebar get-deps
erlexec: HOME must be set
make: *** [deps] Error 1
```

PR #128, opened 2012-03-27
allow either underscores or hyphens in riak-admin command names

vinoski:
The riak-admin command has a number of multi-word commands where some
command words are separated by hyphens and others are separated by
underscores. Many find this inconsistent and confusing. Change the commands
to allow either. Also change the riak-admin usage message to use hyphens
everywhere, as most seem to prefer them. Also change individual command
usage messages to emit whichever form of command the user entered.

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -114,9 +114,9 @@ case "$1" in
         exit 1
         ;;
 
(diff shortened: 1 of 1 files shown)


seancribbs:
:+1: merging

PR #129, opened 2012-04-03
Add {legacy_vnode_routing, false} to default app.config.

jonmeredith:
Set default correctly for new installs.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -31,6 +31,10 @@
               %% single file with both items concatenated together.)
               %{handoff_ssl_options, [{certfile, "/tmp/erlserver.pem"}]},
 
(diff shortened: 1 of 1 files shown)


jtuple:
Change is straightforward. Went ahead and built a 64-bit Ubuntu package off this branch and tested installing it clean versus installing over existing Riak install (and keeping existing config).

On fresh install:

```
(riak@127.0.0.1)2> application:get_env(riak_core, legacy_vnode_routing).
{ok,false}
```

On install-over-existing:

```
(riak@127.0.0.1)2> application:get_env(riak_core, legacy_vnode_routing).
undefined
```

When undefined, Riak defaults to legacy mode. All looks good to me, short of running the `basho_expect` rolling upgrade test, which seems like overkill here.

+1 Merge

PR #134, opened 2012-04-27
Use env to locate bash

sc68cal:
Bash is not always located in `/bin`


slfritchie:
Sean, I think we've already got a PR somewhere in the pipeline to remove bash'isms altogether ... so it isn't clear yet if we'll be using this patch.  Regardless of our decision, thanks for making the patches & submitting the pull request.

PR #137, opened 2012-05-10
riak escript

ghost:
This allows a user to run an escript without having to figure out
where their version of bundled Erlang for Riak is located.

This is extremely useful when we write Erlang scripts to expose
functionality that has not been implemented yet elsewhere, such as
examining message queues.

### DIFF
--- a/rel/files/riak
+++ b/rel/files/riak
@@ -255,11 +255,15 @@ case "$1" in
         fi
         echo "config is OK"
         ;;
+    escript)
+        shift
+        $ERTS_PATH/escript "$@"
+        ;;
     version)
         echo $RIAK_VERSION
         ;;
     *)
-        echo "Usage: $SCRIPT {start|stop|restart|reboot|ping|console|attach|chkconfig|version}"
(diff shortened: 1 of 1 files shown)


bsparrow435:
Tested on Ubuntu and OSX. +1 to merge.

dizzyd:
Hm...is that going to do the right thing when you pass other arguments to the escript? I.e. 

riak escript myescript.erl arg1 arg2

ghost:
That's why I call shift to globber the first argument, and pass in $@ for the rest of the arguments.

```
./bin/riak escript ~/src/basho/basho-escripts/bitcask_dump_keys.escript \
data/bitcask/1118962191081472546749696200048404186924073353216 \
/tmp/

Opening bitcask: "data/bitcask/1118962191081472546749696200048404186924073353216"
Opening output file: "/tmp/1118962191081472546749696200048404186924073353216.keys"

cat /tmp/1118962191081472546749696200048404186924073353216.keys 
__riak_client_test__,key1
```

dizzyd:
Got it, thanks. +1 to merge, but we also need to add this to the EDS version of script.

PR #138, opened 2012-05-10
Extract Protocol Buffers interface from KV

seancribbs:
This is part of a series of pull-requests to extract the Protocol Buffers interface from `riak_kv` and make it available to all applications in an unobtrusive manner.

Related: basho/riak_kv#321

This specific pull-request adjusts app.config to move settings into the new `riak_api`, enables R15B01 support, and temporarily points the `riak_kv` dependency at the branch that includes the refactoring.

The Java and Python clients test suites should pass against this refactoring. Ruby and Erlang clients test suites fail because they directly modify the server or its configuration. Manual testing and benchmarking is encouraged.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -1,6 +1,24 @@
 %% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
 %% ex: ft=erlang ts=4 sw=4 et
 [
+ %% Riak Client APIs config
+ {riak_api, [
+            %% pb_backlog is the maximum length to which the queue of pending
+            %% connections may grow. If set, it must be an integer >= 0.
+            %% By default the value is 5. If you anticipate a huge number of
+            %% connections being initialised *simultaneously*, set this number
+            %% higher.
+            %% {pb_backlog, 64},
+
+            %% pb_ip is the IP address that the Riak Protocol Buffers interface
+            %% will bind to.  If this is undefined, the interface will not run.
+            {pb_ip,   "{{pb_ip}}" },
+
+            %% pb_port is the TCP port that the Riak Protocol Buffers interface
+            %% will bind to
+            {pb_port, {{pb_port}} }
+            ]},
+
  %% Riak Core config
  {riak_core, [
               %% Default location of ringstate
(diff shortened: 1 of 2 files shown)


rebar.config
```diff
@@ -10,7 +10,7 @@
 
 {deps, [
        {cluster_info, ".*", {git, "git://github.com/basho/cluster_info", {branch, "master"}}},
-       {riak_kv, ".*", {git, "git://github.com/basho/riak_kv", {branch, "master"}}},
+       {riak_kv, ".*", {git, "git://github.com/basho/riak_kv", {branch, "sdc-pbc-extract"}}},
```
kellymclaughlin: Just a reminder to make sure this get set back to `master` before merge.

rebar.config
```diff
@@ -1,6 +1,6 @@
 {sub_dirs, ["rel"]}.
 
-{require_otp_vsn, "R14B0[234]"}.
+{require_otp_vsn, "R14B0[234]|R15B01"}.
```
kellymclaughlin: May want to rebase before merging to avoid a conflict since this has been updated on `master` to `R14B0[234]|R15`.

kellymclaughlin:
These changes look good to me,  +1.

rzezeski:
I wonder about the `app.config` changes.  From my understanding when a user upgrades the pkg installers will keep the old app.config.  This means that `riak_api` will revert to defaults if user doesn't migrate PB config from `riak_core` to `riak_api`.  This poses a problem if said user has modified their PB settings (especially IP/Port).

kellymclaughlin:
The other problem is that if they simply use the old version of the config then riak fails to start at all with ugly errors. I made a comment about more helpful error messages in that regard on the `riak_kv` pr. In the interest of backward compatibility maybe `riak_api` should also check for the existence of the properties in the `riak_core` section and use those in the absence of `riak_api` properties in addition to some friendly lager reminders to upgrade the code. Then a couple of versions from now we can take out the `riak_core` check.

rzezeski:
I agree.  I'm going to fix `riak_api` to do just that.  I feel in the future we'll need a better story than "make sure you go into all your app configs and change this stuff or things will break in the future."

rzezeski:
Hopefully this fixes the upgrade path, need to test still.

basho/riak_api#1

rzezeski:
The recent `riak_api` additions fix the upgrade path and allow PB to function with old configs.

+1 to merge.

PR #141, opened 2012-05-21
Remove bashisms from shell scripts which use /bin/sh

### LINKED ISSUE #140: Fix sudo finding under sh
Ubuntu 11 didn't like our switch to `sh`:

https://github.com/basho/riak/commit/2576ef3ec2a8a34e5307a569986374ae7d71c0bf#commitcomment-1356470

We need to fix the method by which we find `sudo` before release.

jaredmorrow:
Fixes #140

`type -p` is not supported by /bin/sh and caused the riak and riak-admin scripts
to think that `sudo` was not available.  In addition `kill -SIGNAL` is not portable
to /bin/sh and should be replaced by `kill -s SIGNAL` instead.

The portability was checked with `checkbashisms` from the devscripts package in
ubuntu 10.04 and was tested on /bin/sh in solaris, ubuntu, and OSX.

### DIFF
--- a/rel/files/riak
+++ b/rel/files/riak
@@ -26,7 +26,7 @@ RIAK_VERSION="git"
 
 # Make sure this script is running as the appropriate user
 if [ "$RUNNER_USER" -a "x$LOGNAME" != "x$RUNNER_USER" ]; then
-    type -p sudo > /dev/null 2>&1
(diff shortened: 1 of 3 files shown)


beerriot:
+1 works for me, thanks

PR #143, opened 2012-05-23
Remove default support for OpenSolaris

jaredmorrow:
OpenSolaris no longer exists as it did when support was added for it. In addition, OmniOS added support
for RIak with the simple patch of s/OpenSolaris/OmniOS/ in the package/Makefile. The OmniOS packages
will still be generated by the OmniOS team, but adding that string to the makefile makes it one step easier for
them to build Riak and that is a Good Thing (tm).


jaredmorrow:
@postwait if you could give a quick review, I'd appreciate it.

PR #147, opened 2012-05-25
Make `ring_creation_size` setting for riak_core more obvious and available

### LINKED ISSUE #139: Ensure ring_creation_size setting is in all app.config files
Not all packages have a default for `ring_creation_size` set in the `app.config` file, so they use the silent default.   Consider a proper default and adding it explicitly to all packages.

jaredmorrow:
Fixes #139


jaredmorrow:
+1'd from @Vagabond and @jtuple in chat

PR #148, opened 2012-05-25
Add staged clustering support to riak-admin

jtuple:
Update `riak-admin` to use the staged clustering support added in response to basho/riak_core#177. The staged cluster interface is provided by `riak-admin cluster` which provides commands to stage a set of changes as well as the plan, commit, and clear commands to view the set of pending changes and commit/clear them.

See the related `basho/riak_core` pull-request: basho/riak_core#181

As part of this change, the non-staged cluster commands are changed to no longer work without adding `-f` such as `riak-admin join -f riak@127.0.0.1`. Otherwise, the commands print-out a message about the new staged clustering system. This change is open to discussion, but I wanted to be opinionated and lightly push users to the new approach as it is much safer and provides much more insight than the non-staged commands.

Typing `riak-admin cluster` details the new clustering commands:

```
Usage: riak-admin cluster <command>

The following commands stage changes to cluster membership. These commands
do not take effect immediately. After staging a set of changes, the staged
plan must be committed to take effect:

   join <node>                    Join node to the cluster containing <node>
   leave                          Have this node leave the cluster and shutdown
   leave <node>                   Have <node> leave the cluster and shutdown

   force-remove <node>            Remove <node> from the cluster without
                                  first handing off data. Designed for
                                  crashed, unrecoverable nodes

   replace <node1> <node2>        Have <node1> transfer all data to <node2>,
                                  and then leave the cluster and shutdown

   force-replace <node1> <node2>  Reassign all partitions owned by <node1> to
                                  <node2> without first handing off data, and
                                  remove <node1> from the cluster.

Staging commands:
   plan                           Display the staged changes to the cluster
   commit                         Commit the staged changes
   clear                          Clear the staged changes
```

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -66,11 +66,118 @@ ERTS_PATH=$RUNNER_BASE_DIR/erts-$ERTS_VSN/bin
 # Setup command to control the node
 NODETOOL="$ERTS_PATH/escript $ERTS_PATH/nodetool $NAME_ARG $COOKIE_ARG"
 
+ensure_node_running()
+{
+    # Make sure the local node IS running
+    RES=`$NODETOOL ping`
+    if [ "$RES" != "pong" ]; then
+        echo "Node is not running!"
+        exit 1
+    fi
+}
+
+cluster_admin()
+{
+    case "$1" in
+        join)
+            if [ $# -ne 2 ]; then
+                echo "Usage: $SCRIPT cluster join <node>"
+                exit 1
+            fi
+            ensure_node_running
+            $NODETOOL rpc riak_kv_console staged_join "$2"
+            ;;
+        leave)
+            if [ $# -eq 1 ]; then
+                node=""
+            elif [ $# -eq 2 ]; then
+                node="$2"
+            else
+                echo "Usage: $SCRIPT cluster leave [<node>]"
+                exit 1
+            fi
+            ensure_node_running
+            $NODETOOL rpc riak_core_console stage_leave "$node"
+            ;;
+        force-remove)
+            if [ $# -ne 2 ]; then
+                echo "Usage: $SCRIPT cluster force-remove <node>"
+                exit 1
+            fi
+            ensure_node_running
+            $NODETOOL rpc riak_core_console stage_remove "$2"
+            ;;
+        replace)
+            if [ $# -ne 3 ]; then
+                echo "Usage: $SCRIPT cluster replace <node1> <node2>"
+                exit 1
+            fi
+            ensure_node_running
+            $NODETOOL rpc riak_core_console stage_replace "$2" "$3"
+            ;;
+        force-replace)
+            if [ $# -ne 3 ]; then
+                echo "Usage: $SCRIPT cluster force-replace <node1> <node2>"
+                exit 1
+            fi
+            ensure_node_running
+            $NODETOOL rpc riak_core_console stage_force_replace "$2" "$3"
+            ;;
+        plan)
+            ensure_node_running
+            $NODETOOL rpc riak_core_console print_staged
+            ;;
+        commit)
+            ensure_node_running
+            $NODETOOL rpc riak_core_console commit_staged
+            ;;
+        clear)
+            ensure_node_running
+            $NODETOOL rpc riak_core_console clear_staged
+            ;;
+        *)
+            echo "\
+Usage: $SCRIPT cluster <command>
+
+The following commands stage changes to cluster membership. These commands
+do not take effect immediately. After staging a set of changes, the staged
+plan must be committed to take effect:
+
+   join <node>                    Join node to the cluster containing <node>
+   leave                          Have this node leave the cluster and shutdown
+   leave <node>                   Have <node> leave the cluster and shutdown
+
+   force-remove <node>            Remove <node> from the cluster without
+                                  first handing off data. Designed for
+                                  crashed, unrecoverable nodes
+
+   replace <node1> <node2>        Have <node1> transfer all data to <node2>,
+                                  and then leave the cluster and shutdown
+
+   force-replace <node1> <node2>  Reassign all partitions owned by <node1> to
+                                  <node2> without first handing off data, and
+                                  remove <node1> from the cluster.
+
+Staging commands:
+   plan                           Display the staged changes to the cluster
+   commit                         Commit the staged changes
+   clear                          Clear the staged changes
+"
+    esac
+}
+
 # Check the first argument for instructions
 case "$1" in
     join)
-        if [ $# -ne 2 ]; then
-            echo "Usage: $SCRIPT join <node>"
+        if [ "$2" != "-f" ]; then
+            echo "The 'join' command has been deprecated in favor of the new "
+            echo "clustering commands provided by '$SCRIPT cluster'. To continue "
+            echo "using the deprecated 'join' command, use 'join -f'"
+            exit 1
+        fi
+
+        if [ $# -ne 3 ]; then
+            echo "Usage: $SCRIPT join -f <node>"
             exit 1
         fi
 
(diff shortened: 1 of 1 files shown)


jaredmorrow:
This needs to be rebased to pick up the underscore/hyphen change that went in this week.

beerriot:
Each command should print a message about what the user should do next.  For example, cluster-join should print something to the effect of, "Run `riak-admin cluster plan` to view staged changes, or `riak-admin cluster clear` to unstage."  

Without a hint like this, it's a bit confusing to see the same output as with the old `riak-admin join <node>`, yet not see the same behavior.  "Okay, now what?"

beerriot:
It looks like the zero-argument cluster-leave function is broken:

```
$ dev/dev2/bin/riak-admin cluster leave
Failed: '' is not a member of the cluster.
```

beerriot:
Is there any way to detect and explain more about why a `commit` may have failed?  The generic "Unable to commit cluster changes. Plan may have changed, please verify the plan and try to commit again" left me guessing.  The particular problem I was having was that not all of my member nodes were up, so I needed to mark the down ones using `riak-admin down <node>` before continuing.

beerriot:
I think I've found a corner case in which `cluster plan` lies.

Start three fresh nodes:

``` shell
$ for i in {1..3}; do dev/dev$i/bin/riak start; done
```

Tell `dev2` to join `dev1`, and plan the migration.

``` shell
$ dev/dev2/bin/riak-admin cluster join dev1@127.0.0.1
Sent join request to dev1@127.0.0.1
$ dev/dev1/bin/riak-admin cluster plan
=============================== Staged Changes ================================
Action         Nodes(s)
-------------------------------------------------------------------------------
join           'dev2@127.0.0.1'
-------------------------------------------------------------------------------


NOTE: Applying these changes will result in 1 cluster transition

###############################################################################
                         After cluster transition 1/1
###############################################################################

================================= Membership ==================================
Status     Ring    Pending    Node
-------------------------------------------------------------------------------
valid     100.0%     50.0%    'dev1@127.0.0.1'
valid       0.0%     50.0%    'dev2@127.0.0.1'
-------------------------------------------------------------------------------
Valid:2 / Leaving:0 / Exiting:0 / Joining:0 / Down:0

WARNING: Not all replicas will be on distinct nodes

Transfers resulting from cluster changes: 32
  32 transfers from 'dev1@127.0.0.1' to 'dev2@127.0.0.1'
```

Now kill `dev2` (who knows why, it could happen), and then commit the migration plan:

``` shell
$ dev/dev2/bin/riak stop
ok
$ dev/dev1/bin/riak-admin cluster commit
Cluster changes committed
```

Everything will just sit in Pending as long as we leave `dev2` down.

``` shell
$ dev/dev1/bin/riak-admin member_status
================================= Membership ==================================
Status     Ring    Pending    Node
-------------------------------------------------------------------------------
valid     100.0%     50.0%    'dev1@127.0.0.1'
valid       0.0%     50.0%    'dev2@127.0.0.1'
-------------------------------------------------------------------------------
Valid:2 / Leaving:0 / Exiting:0 / Joining:0 / Down:0
```

But let's add `dev3` anyway…

``` shell
$ dev/dev3/bin/riak-admin cluster join dev1@127.0.0.1
Sent join request to dev1@127.0.0.1
$ dev/dev1/bin/riak-admin cluster plan
=============================== Staged Changes ================================
Action         Nodes(s)
-------------------------------------------------------------------------------
join           'dev3@127.0.0.1'
-------------------------------------------------------------------------------


NOTE: Applying these changes will result in 2 cluster transitions

###############################################################################
                         After cluster transition 1/2
###############################################################################

================================= Membership ==================================
Status     Ring    Pending    Node
-------------------------------------------------------------------------------
valid     100.0%     50.0%    'dev1@127.0.0.1'
valid       0.0%     50.0%    'dev2@127.0.0.1'
valid       0.0%      0.0%    'dev3@127.0.0.1'
-------------------------------------------------------------------------------
Valid:3 / Leaving:0 / Exiting:0 / Joining:0 / Down:0

WARNING: Not all replicas will be on distinct nodes

Transfers resulting from cluster changes: 32
  32 transfers from 'dev1@127.0.0.1' to 'dev2@127.0.0.1'

###############################################################################
                         After cluster transition 2/2
###############################################################################

================================= Membership ==================================
Status     Ring    Pending    Node
-------------------------------------------------------------------------------
valid      50.0%     34.4%    'dev1@127.0.0.1'
valid      50.0%     32.8%    'dev2@127.0.0.1'
valid       0.0%     32.8%    'dev3@127.0.0.1'
-------------------------------------------------------------------------------
Valid:3 / Leaving:0 / Exiting:0 / Joining:0 / Down:0

WARNING: Not all replicas will be on distinct nodes

Transfers resulting from cluster changes: 21
  10 transfers from 'dev1@127.0.0.1' to 'dev3@127.0.0.1'
  11 transfers from 'dev2@127.0.0.1' to 'dev3@127.0.0.1'

$ dev/dev1/bin/riak-admin cluster commit
Unable to commit cluster changes. Plan may have changed, please verify the plan and try to commit again
```

So we see that the earlier migration is still pending, and now we have a new plan staged, but we can't commit it because `dev2` is still down.  But, if we mark `dev2` as down, we can commit the migration…

``` shell
$ dev/dev1/bin/riak-admin down dev2@127.0.0.1
Success: "dev2@127.0.0.1" marked as down
```

…but let's check the plan again first:

``` shell

$ dev/dev1/bin/riak-admin cluster plan
=============================== Staged Changes ================================
Action         Nodes(s)
-------------------------------------------------------------------------------
join           'dev3@127.0.0.1'
-------------------------------------------------------------------------------


NOTE: Applying these changes will result in 0 cluster transitions
```

I think this plan is a lie, because I can now commit it, and transfers will, indeed, happen:

``` shell
$ dev/dev1/bin/riak-admin cluster commit
Cluster changes committed
$ dev/dev1/bin/riak-admin member_status
================================= Membership ==================================
Status     Ring    Pending    Node
-------------------------------------------------------------------------------
down        0.0%      --      'dev2@127.0.0.1'
valid      67.2%      --      'dev1@127.0.0.1'
valid      32.8%      --      'dev3@127.0.0.1'
-------------------------------------------------------------------------------
Valid:2 / Leaving:0 / Exiting:0 / Joining:0 / Down:1
```

Or, at least `member_status` thinks they happened.  This could be the lie instead.

If we bring `dev2` back up, the first transfer completes as well.

``` shell
$ dev/dev2/bin/riak start

$ dev/dev1/bin/riak-admin member_status
================================= Membership ==================================
Status     Ring    Pending    Node
-------------------------------------------------------------------------------
valid      34.4%      --      'dev1@127.0.0.1'
valid      32.8%      --      'dev2@127.0.0.1'
valid      32.8%      --      'dev3@127.0.0.1'
-------------------------------------------------------------------------------
Valid:3 / Leaving:0 / Exiting:0 / Joining:0 / Down:0
```

So, I don't know that there's anything broken, but there seems to be something confused, which I thought I'd point out, in case it's indicative of some other issue.

Vagabond:
I ran into the same things as Bryan, so there's certainly something here that needs to be addressed.

Vagabond:
Oh, and making the error messages for when a commit fails useful, like we talked about, would be great.

jtuple:
I've pushed out new code to both basho/riak#148 and basho/riak_core#181 that should address raised concerns plus fix some bugs that I ran into when testing further. It would be great if you could both could re-test things. The plan should never lie. Also, you can no longer plan/commit while transfers are on-going. To solve this, I've added `riak-admin transfer_limit`. Users can set the transfer limit to 0, wait for the ring to stabilize, plan/commit new changes, then increase the transfer limit. A little annoying, but makes things entirely safe.

One thing to note is that any node that is ever downed with `riak-admin down` will always immediately re-join the cluster and claim partitions whenever it comes online. Even if the node was staged for joining, but downed before actually committing the join. There is no easy way around this without major changes to the claimant code. I'm willing to live with this for now.

Also, the 'down rejoining the cluster' behavior never shows up in any plan.  It doesn't make sense. If 4 different nodes are in the down state, how each would claim would be a property of which nodes came back online and at what time. It's entirely non-deterministic and based on outside behavior (nodes coming back online, etc). In short, there's a reason it's still `riak-admin down` and not `riak-admin cluster down`: `down` lives outside of the staged clustering system.

With that said, the reported down bug in this review has been fixed. Downing nodes should never make the plan lie about what will happen with any staged changes to non-down nodes.

Vagabond:
I'm +1 on this part of the code now.

PR #151, opened 2012-05-26
Treatment for commented lines in node.args

shairontoledo:
Hi, I was getting an error using riak-admin commands, it was printing 

"Node is not running!"

So I took a day to discover that only the last line matched with "-name" was taken by riak-admin script, like "foobar@..." in this context

-name riak@10.32.218.65
# -name riak@127.0.0.1
# -name foobar@127.0.0.1

The patch removes out lines started with #.

Thx

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -32,7 +32,7 @@ fi
 cd $RUNNER_BASE_DIR
 
 # Extract the target node name from node.args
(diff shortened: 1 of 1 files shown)


seancribbs:
Why not just put the `^` at the front of the original regexp, which accomplishes the same thing?

shairontoledo:
No reason to verbose mode, consider using `^` in the beginning of the line.

PR #153, opened 2012-05-30
Support FreeBSD packaging

jaredmorrow:
Create a package that can be installed on freebsd using `pkg_add <pkg>`


jaredmorrow:
Merging with known problems surrounding pkg_delete and the package.   Future commits will change directory structure and hopefully quiet down pkg_delete.

PR #154, opened 2012-05-31
print node name and IP addresss to stdout when starting

joecaswell:
per customer request


jaredmorrow:
Is this still a requirement by that customer?  I question whether outputting something new over stdout on start would mess up deployment scripts used in the wild.  It might be a more impactful change than we expect.

jaredmorrow:
Closing for now, will reopen if this comes up again as a customer request.

PR #156, opened 2012-06-06
Fix dependency on tags being available on every repository

jaredmorrow:
Package build passed on all platforms on buildbot.


jaredmorrow:
Tested on EDS, so going to merge so we can have working packages again

PR #161, opened 2012-06-13
'riak attach -f': Enable attaching to the node without pinging it

ghost:
Right now, if the node is unresponsive to `riak ping`, `riak attach` will not work. 

There are times when the node is still running, but unresponsive to pings. In these situations, it is useful to run diagnostics or attempt to cleanly shut down the node via the Erlang console.

This adds a `-f` option to `riak attach`, bypassing pinging the node.

### DIFF
--- a/rel/files/riak
+++ b/rel/files/riak
@@ -210,7 +210,7 @@ case "$1" in
         ;;
 
     attach)
-        if [[ $2 == "-f" ]]; then
+        if [ "$2" = "-f" ]; then
           echo "Forcing connection..."
         else  
           # Make sure a node is running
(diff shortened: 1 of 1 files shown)


rel/files/riak
```diff
@@ -210,12 +210,16 @@ case "$1" in
         ;;
 
     attach)
-        # Make sure a node IS running
-        RES=`ping_node`
-        ES=$?
-        if [ "$ES" -ne 0 ]; then
-            echo "Node is not running!"
-            exit $ES
+        if [[ $2 == "-f" ]]; then
```
jonmeredith: I think [[ is a bashism - how about if 

```
test "$2" == "-f"
```

 instead?

PR #165, opened 2012-06-14
Change ownership of /etc/riak for RPM packages

dcarley:
The current RPM packages set the ownership of `/etc/riak` and the configurations files therein to `riak:riak`. This is typically considered bad practice because the daemon shouldn't have, or need, permissions to modify it's own configs.

The first commit in this PR rectifies that. The default file ownership is set to `root:root` and log/state directories are then explicitly owned to `riak:riak`. This matches the behaviour of the Debian packages.

The second commit omits some attributes that already come from the `%install` section, thus simplifying the SPEC. It could be squashed into the prior commit if preferred.


jaredmorrow:
@dcarley thanks for fixing this oversight.   I built a package and tested against our forthcoming 1.2 release.   I will add you to the THANKS file.

dcarley:
Thanks!

PR #169, opened 2012-06-19
Allow commented out "-name" lines.

jamesproud:
Example vm.args:
# -name riak@127.0.0.1

-name riak@127.0.0.2

riak-admin concatenates the above as "#-name riak@127.0.0.1 -name riak@127.0.0.2" currently.

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -37,7 +37,7 @@ fi
 cd $RUNNER_BASE_DIR
 
 # Extract the target node name from node.args
-NAME_ARG=`grep '\-[s]*name' $RUNNER_ETC_DIR/vm.args`
+NAME_ARG=`grep '^[^#]*\-[s]*name' $RUNNER_ETC_DIR/vm.args`
 if [ -z "$NAME_ARG" ]; then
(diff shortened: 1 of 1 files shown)


n00k:
shouldn't the regex be '^-[s]*name', so it will match 
-name riak@node.host
but not a commented name like
#-name test-riak@node.hode
or a comment about the switch like
#use the -name switch to specify the name of the node
or another switch that happens to contain that like
-administrator-name admin

seancribbs:
Also note that `[s]*name` is wrong and should be `s?name`. Erlang will ignore a switch like `-sssssssssname`. I vote for `grep "^[[:space:]]*-s\?name"`

bookshelfdave:
I tested all cases mentioned above on OSX and the regex `grep "^[[:space:]]*-s\?name"` works. We'll need to test on other platforms as well.

bookshelfdave:
Tested on OSX, Ubuntu 12.04 x64, FreeBSD 9 x64

bookshelfdave:
My apologies, I recreated this fix in branch dip_gh169 as I was unsure about my rebase/squash.

Please see https://github.com/basho/riak/pull/175

PR #174, opened 2012-06-25
Install freebsd files to proper directories based on hier(7)

jaredmorrow:
Change where we install files on FreeBSD based on the layout given by http://www.freebsd.org/cgi/man.cgi?query=hier manpage.

The layout is as follows:
- `platform_bin_dir` = `/usr/local/sbin`
- `platform_data_dir` = `/var/db/riak`
- `platform_etc_dir` = `/usr/local/etc/riak`
- `platform_lib_dir` = `/usr/local/lib/riak`
- `platform_log_dir` = `/var/log/riak`
- man pages = `/usr/local/man`

The man pages and `platform_etc_dir` should have root:wheel ownership while everything else should have riak:riak ownership.

### DIFF
--- a/package/freebsd/+DISPLAY
+++ b/package/freebsd/+DISPLAY
@@ -1,17 +1,20 @@
 Thank you for installing Riak.
 
-Riak has been installed in /usr/local/riak owned by user:group riak:riak
+Riak has been installed in /usr/local owned by user:group riak:riak
 
 The primary directories are:
 
-    {platform_bin_dir, "/usr/local/riak/bin"}
-    {platform_data_dir, "/usr/local/riak/data"}
-    {platform_etc_dir, "/usr/local/riak/etc"}
-    {platform_lib_dir, "/usr/local/riak/lib"}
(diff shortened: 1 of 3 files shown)


bookshelfdave:
Verified permissions on riak, riak-admin, search-cmd, man pages. Ran basho_bench against the instance, and the data dir was populated. Verified with Jared that config files in /usr/local/etc/riak should be root:wheel.

Log files show up in the right place. 

+1

sc68cal:
:beer:

jaredmorrow:
@sc68cal no problem, let me know if you have any other issues/requests.  I don't have a lot of go-to FreeBSD users, so if you use it please give me feedback.   The obvious missing piece is some sort of startup/init.d script, but that won't make it for 1.2.

sc68cal:
I'll give this a shot on my -CURRENT box and report back.

I have a little bit of experience writing startup scripts for [rc.conf(5)](http://www.freebsd.org/cgi/man.cgi?query=rc.conf&sektion=5) - so I may be able to help with that.

PR #175, opened 2012-06-26
Change regex to allow commented out -name lines

bookshelfdave:
The regex should ignore vm.args lines like the following:
# -name riak@127.0.0.1
# -name test-riak@node.hode
# use the -name switch to specify the name of the node

-administrator-name admin
-sssssssssname foo


jaredmorrow:
This does not work on Solaris 10, I will investigate how to change it to work properly.

PR #187, opened 2012-07-18
Create SMF service for epmd on SmartOS

jaredmorrow:
Riak is started and stopped on SmartOS using the [SMF system](http://wiki.joyent.com/display/smart/About+the+Service+Management+Facility). This system keeps track of every process started by a service.

This was causing difficulties with [epmd](http://www.erlang.org/doc/man/epmd.html) because that process would not shut down on `svcadm disable riak` causing Riak to go into a "maintenance" mode.

This pull request creates a separate SMF service for epmd called "riak-epmd".  This will do its best to start and stop epmd in a sane manner, allowing for Riak to be started and stopped also in sane way.

If an epmd process is active, but with no registered names, it is killed and restarted.

See: basho/riak_ee#75

### DIFF
--- a/package/smartos/+DISPLAY
+++ b/package/smartos/+DISPLAY
@@ -22,9 +22,18 @@ Man pages are available for riak(1), riak-admin(1), and search-cmd(1)
 This package is SMF enabled, which means you can use SMF to 'enable',
 'disable' or 'restart' the persistent daemon process, e.g.:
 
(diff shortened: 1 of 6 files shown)


jaredmorrow:
Merging based on riak_ee review.

bahamas10:
awesome, excited for this.

PR #193, opened 2012-07-26
Declare that the 'folsom' app should be loaded, not started at boot time

russelldb:
When a release was generated with r14 the folsom app was started at
boot time despite being an included app of riak_core. Setting the app_type()
to `load` in the reltool.config fixes this.

Tested on r14 and r15.

Also remove redundant app.config vars for stats.


jonmeredith:
+1 merge, starts under r14b04 and r15b01 now.

Issue #192, opened 2012-07-25
Permissions not checked on $PIPE_DIR in riak script


bsparrow435:
If the /tmp/riak/ directory is owned by root user, riak start will fail but riak console will succeed. Add a permissions check on this directory.  Will attach PR shortly.

bsparrow435:
https://github.com/basho/riak/pull/194

Looking for comments.

jaredmorrow:
Fixed in 1.3 and master (1.4)

PR #194, opened 2012-07-26
Added writable check for PIPE_DIR

bsparrow435:
Re: Issue 192.

In certain situations PIPE_DIR exists and is owned by a different user.  This will silently crash riak start but riak console will succeed.


jaredmorrow:
+1, I will merge and bring it into 1.3.

Fixes #192

PR #196, opened 2012-07-30
Make dev targets in Makefile depend on all so running devrel or stagedevrel from a fresh checkout works.

kellymclaughlin:


joedevivo:
+1

PR #206, opened 2012-08-16
update rebar to 2.0.0

bookshelfdave:


jaredmorrow:
```
WARNING: deprecated app option used
Option 'app' has been deprecated
in favor of 'apps'.
'app' will be removed soon.
```

Some targets used in the makefile use outdated commands.  Running through the various Makefile targets to see what is deprecated or not working would be a good thing to do.

jtuple:
It would still be a good idea to double-check if there were any remaining deprecated targets. However, now that basho/erlang_js#29 has been merged, top-level Riak can no longer compile. I've tested that this pull-request fixes that problem. We should go ahead and merge this now to get Riak compiling again, and fix any other deprecated usage down the road if they're discovered.

+1 merge

PR #207, opened 2012-08-29
a fix for 'make test'

bookshelfdave:


beerriot:
+1 this made "make test" run tests again for me, instead of saying there were no tests to run

PR #208, opened 2012-08-31
Adding the debian distro codename to the version metadata.

jsmartin:
Necessary for proper apt repo creation.


jaredmorrow:
This breaks in 32 bit ubuntu  https://gist.github.com/2d99c838fc4fa9b4b42c

jsmartin:
Should be good now.  Found the proper way to determine the architecture dpkg uses to construct filenames.

jaredmorrow:
+1, nice work.

PR #212, opened 2012-09-10
cluster_info removal

evanmcc:
Remove cluster info from the release and applications stuff.
Remove the call from riak-admin.

This change also relies upon:
https://github.com/basho/riak_kv/pull/393
https://github.com/basho/riak_search/pull/130
https://github.com/basho/riak_core/pull/226
https://github.com/basho/riak_pipe/pull/54


jaredmorrow:
This removal would be with the assumption that riaknostic will be included by default, correct?

evanmcc:
Kind of.  The idea was that cluster info would join riaknostic in our package of optional diagnostic tools, but since then we've come to the conclusion that it would go into riaknostic, since too many customer have complained about not being able to use riaknostic, since they can't install additional stuff in their security-restricted environments.

jaredmorrow:
@evanmcc I'm going to assume this is going to be post-1.4 considering none of the other PR's are closed yet?

evanmcc:
Discussion is fitfully ongoing, but I suspect that due to done
environmental restrictions this will be closed and discarded soon.

PR #213, opened 2012-09-12
Raise the ETS table limit

rzezeski:
Raise the ETS table limit to give headroom for Riak Search if it is
enabled.


jonmeredith:
+1 merge

PR #216, opened 2012-09-17
doc/architecture.txt: s/=</<=/g;

jjuran:
Replace =< with <= (less than or equal to).


jonmeredith:
Erlangisms like to creep in everywhere

```

(jdm3@jons-macpro.jonmeredith.com)2>  1 =< 2.
true
(jdm3@jons-macpro.jonmeredith.com)3>  1 <= 2.
* 1: syntax error before: '<='
```

PR #217, opened 2012-09-17
doc/architecture.txt:  "compare TWO versions"

jjuran:
Write "two", not "to".


PharkMillups:
+1. Thanks for the fix.

PR #219, opened 2012-09-17
doc/basic-setup.txt:  Strike out-of-place comma

jjuran:
Remove the comma between the indirect object ("the cluster") and the
direct object ("up and running").


jaredmorrow:
Thanks for the contribution.

PR #220, opened 2012-09-17
doc/raw-http-howto.txt:  Strike out-of-place comma

jjuran:
Remove the comma in "key-level, document requests".  A comma separates
two modifiers of the same term, but "document" binds tighter, modifying
"requests" itself, whereas "key-level" modifies the compound term
"document requests" -- so a comma is no more appropriate here than it
would be in "striped candy cane".


jaredmorrow:
Thanks for the contribution.

PR #221, opened 2012-09-28
Add configuration option for 2i backpressure.

kellymclaughlin:
Corresponds to [this](https://github.com/basho/riak_kv/pull/399) pr to add backpressure for 2I operations.


beerriot:
The choice of whether or not to use backpressure in riak_kv#399 is controlled by core capability negotiation, not by application environment. It will enable itself automatically as soon as every node in the cluster supports it. I see no reason to provide the option to disable it, so I think the change in this PR is unecessary.

PR #222, opened 2012-10-02
Fix the link to Ruby client in README

Domon:


PharkMillups:
+1. Thanks for the fix

PR #223, opened 2012-10-02
Don't use lager console logging for release builds

kellymclaughlin:
Avoid redundant and potentially expsensive console logging in release builds. Move the lager handler configuration to be a template substitution so that console logging is still the default for devrel builds


jaredmorrow:
+1

PR #224, opened 2012-10-08
Prevent Riak from starting if ports used in app.config are not unique

scalp42:
Related to issue : https://github.com/basho/riak_core/issues/238


evanmcc:
Thanks for the effort, but we're going to go with a check that fits into our model a bit better.  PR to come soon.

PR #225, opened 2012-10-09
[WIP] Joyent updated their SmartOS datasets, verify pkgsrc repo to determine d...

scalp42:
...ataset used and import correct manifest.xml

`http://wiki.joyent.com/wiki/display/jpc2/Available+Joyent+Public+Cloud+Machine+Images#AvailableJoyentPublicCloudMachineImages-GeneralPurposeSmartMachines`

If you use a SmartOS machine with `SmartOS`  dataset, you will end up with gcc46 with the path `/opt/local/lib/amd64/libgcc_s.so.1`

If you use a SmartOS machine with `Base` or `Standard`, you will end up with gcc47 with the path `/opt/local/lib/amd64/libgcc_s.so.1`

Even if you use an old dataset and change yourself the pkgin repo to point to latest Joyent one, you will end up with gcc47 in the path.

Also, `whoami` moved from `ucb` path to `/opt/local/gnu/bin/`.


scalp42:
Riak SmartOS package can't start (even with those 2 patches) on 2012Q2 repo :

  [root@prod-riak5 /var/svc/log]# riak start
  Attempting to restart script through sudo -H -u riak

---

  Warning: please use 'svcadm enable riak' instead

---

  ld.so.1: beam.smp: fatal: libncurses.so.5: open failed: No such file or directory
  /opt/local/sbin/riak: line 97: 21662: Killed
  ld.so.1: beam.smp: fatal: libncurses.so.5: open failed: No such file or directory
  /opt/local/sbin/riak: line 130: 21669: Killed
  Error reading /opt/local/etc/riak/app.config

No9:
Has a similar problem on openindiana erlang was expecting libncurses.so.5 in /usr/lib 

```
# ln -s /usr/gnu/lib/libncurses.so.5 /usr/lib/libncurses.so.5
```

N.B. YMMV

scalp42:
Thanks @No9 will look into it

jaredmorrow:
We found these issues as well and were fixed for the Riak 1.2 release.  Thanks for the PR!

scalp42:
Sweet!

PR #226, opened 2012-10-09
add port uniqueness checking

evanmcc:
add a check to nodetool's chkconfig mode to make sure that all of our ports are unique.

@scalp42 this should fix your issue when it goes in.

@jaredmorrow How does this look?

### DIFF
--- a/rel/files/nodetool
+++ b/rel/files/nodetool
@@ -7,6 +7,7 @@
 %% nodetool: Helper Script for interacting with live nodes
 %%
 %% -------------------------------------------------------------------
+-mode(compile).
 
 main(Args) ->
     ok = start_epmd(),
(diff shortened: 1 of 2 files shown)


jaredmorrow:
## To Test

### Initial test
- First do a `make rel` to get a working riak install
- Test that `rel/riak/bin/riak [start, chkconfig, console]` all work with a stock config

### app.config errors to induce

Do each of the following to a stock `app.config`
1. Add a comma / period somewhere to break parsing
   - 'riak chkconfig' should **fail**
2. Set `pb_port, handoff_port, http_port, https_port` in any combination to have identical ports
   - 'riak chkconfig' should **fail**
3. Remove `pb_port` setting
   - 'riak chkconfig' should put out a warning and **pass**
   - 'riak start' should warn on stderr and start normally
4. Remove `handoff_port` setting
   - `riak chkconfig` should **fail**

PR #228, opened 2012-10-10
Adding support for using bloom filters in mapreduce jobs

whitenode:
I have created a solution that allows bloom filters to be created, updated and used for filtering as part of mapreduce jobs. Part of the solution is a small OTP application that manages local caching of filters during map phase filtering as well as configuring the bucket used for filter storage on start-up. I have updated the rebar.config and reltool.config files to add the dependency on the riakbloom application and ensure this is started up together with Riak. Is this something you would consider integrating into Riak?

Best regards,

Christian

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -13,5 +13,6 @@
        {riak_kv, ".*", {git, "git://github.com/basho/riak_kv", {branch, "master"}}},
        {riak_search, ".*", {git, "git://github.com/basho/riak_search",
                                 {branch, "master"}}},
-       {riak_control, ".*", {git, "git://github.com/basho/riak_control", {branch, "master"}}}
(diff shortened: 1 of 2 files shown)


jaredmorrow:
@whitenode after discussing this a bit, we decided right now we don't have the bandwidth to maintain and support another external dependency to Riak.  

However, this is exactly the thing we encourage our community to do and we have an entire page dedicated to community contributions.  We appreciate any and all community contributions, so please consider adding your project to [our community page](http://docs.basho.com/riak/latest/references/Community-Developed-Libraries-and-Projects/).  You can put in a PR here https://github.com/basho/basho_docs to modify that page if that is something you are interested in doing.  

Also, I think people on the [Riak Users mailing list](http://lists.basho.com/mailman/listinfo/riak-users_lists.basho.com) would be interested as well, so please be sure to post something there.  Thanks once again for the contribution.

PR #231, opened 2012-10-16
Added `riak-admin cluster members`

lenary:
This lets users see current cluster membership. It does exactly the
same as `riak-admin member-status` only it's in the cluster subcommand,
where I feel it should live.

I kept tripping up on the fact that the cluster membership commands weren't in the same place as the cluster altering commands, so I decided to make it so! Open Source ftw!

The old commands still work, and perhaps there's a better way of keeping backwards compatibility than duplicating code. I guess one can just call the other.

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -148,6 +148,22 @@ cluster_admin()
             ensure_node_running
             $NODETOOL rpc riak_core_console clear_staged
             ;;
+        members)
+            if [ $# -ne 1 ]; then
+                echo "Usage: $SCRIPT cluster $1"
+                exit 1
+            fi
+
+            # Make sure the local node IS running
+            RES=`$NODETOOL ping`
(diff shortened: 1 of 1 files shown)


jaredmorrow:
I like the idea, but might not get a chance to look at it in detail for a little bit.  Sorry to push it off, just want to make sure you know someone saw this!  Thanks for the PR!

jaredmorrow:
Since @jtuple did the work on the new cluster subcommand I'll let him decide if he wants this functionality moved.

slfritchie:
Ping ... is this PR needed for Riak 2.0?

jrwest:
It is not needed. At a minimum it needs to be re-opened to point at `develop` instead of `master` and I imagine requires a merge/rebase given the changes to `riak-admin` this time around. Closing for now with the expectation that @lenary will re-open if he wants.

Issue #232, opened 2012-10-20
Debian package init script does not properly exit on `status` command


jaredmorrow:
When issuing a `/etc/init.d/riak status` command, if `riak` is not running, the exit status of the command should be a positive integer rather than zero.

PR #233, opened 2012-10-22
Properly exit on `/etc/init.d/riak status` command

jaredmorrow:
When running `/etc/init.d/riak status`, if the status fails
(Riak isn't running) not only should a message print out
that states that fact, but also the exit status should be non-zero.

This change addresses issue basho/riak#232


jaredmorrow:
Merge based on basho/riak_ee#96 code review.

PR #236, opened 2012-11-09
Allow dynamic number of devrel nodes to be created

evanmcc:
Allow for any number of devrel / stagedevrel nodes to be created.
`make devrel DEVNODES=60` will create a 60 node devrel

Original PR basho/riak_ee#90.


jaredmorrow:
Shouldn't the dev*_vars_config files be gone in this PR?  Also, SNMP is only a EE feature so make sure that port is left out to avoid confusion.

evanmcc:
I think that those pushes take care of it.

jaredmorrow:
+1, making note that you intended to bump up the node count by default from 4 to 6

I'm fine with either

evanmcc:
Good catch, will revert.

PR #237, opened 2012-11-12
Fix stagedevrel, fix hanging close paren.

evanmcc:
as pointed out my @beerriot, stagedevrel was following the devrel code path.


beerriot:
it works! +1, thanks

PR #239, opened 2012-11-14
Remove legacy listkeys config option

kellymclaughlin:
Legacy keylisting is being removed from riak_kv so the `legacy_keylisting` configuration option is now obsolete.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -148,13 +148,6 @@
             %% Only set true if *all* nodes in the cluster are upgraded to 1.0
             {vnode_vclocks, true},
 
-            %% This option enables compatability of bucket and key listing
-            %% with 0.14 and earlier versions. Once a rolling upgrade to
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
Related `riak_kv` pr is [here](https://github.com/basho/riak_kv/pull/400).

slfritchie:
Hrm, I thought that I saw this same app config change with the Luke app removal review that I did recently?  If I'm thinking of another config item that @beerriot removed, then great.

This same packaging change should also be applied to the riak_ee repo?

I don't see anything in https://github.com/basho/riak_kv/pull/436 that looks objectionable, +1 to merge.

kellymclaughlin:
I think it was a similar, but different item that was in the luke removal. I've added a pr [here](https://github.com/basho/riak_ee/pull/103) to mirror this change in `riak_ee`.

PR #241, opened 2012-11-14
Make devrel and stagedevrel make commands more portable

jaredmorrow:
The shell command 'seq' is not available on some *BSD systems
including several OSX flavors.  Instead this uses 'awk'
to perform the same task.  In addition the `gen_dev` script
was changed to use bin/sh rather than bin/bash.


evanmcc:
+1

PR #244, opened 2012-11-16
Add riaknostic to riak builds

bsparrow435:
This adds riaknostic to the rebar.config and adds make targets for riaknostic so it is grabbed, built, and correctly copied/linked for rel/devrel/stage builds.


jaredmorrow:
I'm looking at it now, but for some reason 'make' is not being smart about not running 'deps' for each dev/dev\* on a devrel so 'make devrel' is taking a long time.  I will work on it and will commit to the branch.

PR #248, opened 2012-11-21
Set riak_sysmon's gc_ms_limit default value to zero

slfritchie:
Customer feedback during the Riak 1.2 release cycle has
demonstrated that the R15 virtual machine can have serious
mutex contention problems around the time-of-day clock when
Riak is under heavy load on "large" multi-core systems,
e.g. 24 core machines.  Setting the default value to zero
will disable the 'long_gc' system monitor messages.


rzezeski:
+1 to merge

PR #251, opened 2012-12-05
`riak stop` does not behave properly on BSD systems

jaredmorrow:
When using `riak stop` in BSD systems a regular expression was failing that attempted to capture the PID of the running Riak node.  This PID was used in the `riak stop` command to properly wait for a node to stop completely before returning.  

The user experience resulted in the following situation:

```
$ ./bin/riak start
$ ./bin/riak ping 
pong
$ ./bin/riak stop
ok
$ ./bin/riak ping
pong
$ ./bin/riak ping
pong
....  some time
$ ./bin/riak ping
pong
$ ./bin/riak ping
Node 'dev1@127.0.0.1' not responding to pings.
```

The expected behavior is like that on Linux, `riak stop` should not return until Riak is fully stopped.

### DIFF
--- a/rel/files/nodetool
+++ b/rel/files/nodetool
@@ -59,6 +59,8 @@ main(Args) ->
     end,
 
     case RestArgs of
+        ["getpid"] ->
+            io:format("~p\n", [list_to_integer(rpc:call(TargetNode, os, getpid, []))]);
         ["ping"] ->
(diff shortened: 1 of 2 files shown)


jaredmorrow:
Test added to basho/riak_test here basho/riak_test#116

joedevivo:
Not only do I like it, but it totally works via riak_test and manual devrel testing. +1

PR #257, opened 2012-12-13
Add syslog support for lager to riak

Vagabond:
Provide the necessary dependencies so that you can enable syslog as a lager backend for riak.

syslog and lager_syslog are included in the release, but not started by default because we don't want to unnecessarily load the port driver if we aren't going to use it.

To enable syslog support, add this under lager's section of the app.config

```
{handlers, [
                     {lager_syslog_backend, ["riak", daemon, info]},
                      ....     
                   ]},
```

That will make lager log any info messages (or above) to the daemon facility of syslog with the "riak" identity string.

Along with this, the code for https://github.com/basho/lager_syslog should probably be vetted, as we've never shipped it before.

lager_syslog is compatible with both lager 1.0.x, 1.2.x and 2.0.x (the upcoming lager release). As part of 1.3, we can pin the lager_syslog (and/or the lager) dependency to the 1.2.1 tag, instead of the current dependency on master.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -9,6 +9,7 @@
 {erl_opts, [debug_info, fail_on_warning]}.
 
 {deps, [
+       {lager_syslog, ".*", {git, "git://github.com/basho/lager_syslog", {branch, "master"}}},
        {cluster_info, ".*", {git, "git://github.com/basho/cluster_info", {branch, "master"}}},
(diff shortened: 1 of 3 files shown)


jaredmorrow:
I think a commented out `lager_syslog_backend` example would be a nice addition to the `app.config`

jaredmorrow:
+1, tests pass and I see no issues with the lager_syslog code.

Need to patch riak_ee as well.

PR #259, opened 2012-12-18
Remove unused bits of legacy-mapred configuration

beerriot:
The configuration settings `mapred_queue_dir`, `mapper_batch_size`, and `map_cache_size` are not used now that the legacy (Luke-based) MapReduce system has been removed. Remove them from config files to avoid confusion.


jaredmorrow:
:+1:  goodbye legacy M/R

PR #265, opened 2013-01-08
Add daemontools support

doubleyou:
At Mochi Media, we are starting Riak using 'riak console' instead of 'riak start', since daemontools doesn't work with background applications, and run_erl doesn't properly handle signals.

Also, we had to implement our own command "riak remsh", which is basically "erl -remsh".
- Add 'riak dt_start' command to be executed from daemontools run script.
- Extend 'riak attach' command to determine riak start mode and use
  "erl remsh" for daemontools-started node.

P.S. If this goes upstream, we'll finally be able to use 100% non-customized Riak :)

### DIFF
--- a/rel/files/riak
+++ b/rel/files/riak
@@ -242,7 +242,15 @@ case "$1" in
         fi
 
         shift
-        exec $ERTS_PATH/to_erl $PIPE_DIR
+        RES=`$NODETOOL rpcterms os cmd "\"echo \\$RIAK_STARTING_MODE\"." < /dev/null`
+        if [[ "$RES" =~ "daemontools" ]]; then
+            UUID=`uuidgen`
+            REMSH_NODE=`echo $NAME_ARG | sed "s/-name \([^\@]*\)/riak-attach-"$UUID"/1"`
+            echo "-name $REMSH_NODE"
+            exec $ERTS_PATH/erl -name $REMSH_NODE $COOKIE_ARG
+        else  
+            exec $ERTS_PATH/to_erl $PIPE_DIR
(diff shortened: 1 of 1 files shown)


jaredmorrow:
Thanks for the PR @doubleyou, we've talked about doing something similar actually with the remsh because of our problems with `attach` you mentioned.

We are in a release cycle right now, so I'll have to come back to this PR after 1.3 branches for RC status.

Thanks again.

doubleyou:
Frankly, run_erl way sucks in many ways.
- Too easy to accidentally close Erlang, by pressing usual Ctrl+C instead of expected Ctrl+D.
- Attached console is extremely inconvenient, compared to the one from "erl -remsh"
- All attached console IO goes into logs as well, might be bad for logs processing software
- Its logs rotation ways is inconvenient

Have you ever considered getting rid of the run_erl completely? Lager can do logs rotation much better. And multilog from daemontools does this even more awesome, IMO, (I'm even thinking of making another logs rotation method for lager, completely similar to multilog's way).

Besides logs rotation, I don't see any reasons for keeping run_erl, at least so far I see its purpose.

jaredmorrow:
This is related to #254

jaredmorrow:
@doubleyou is the change made in basho/riak#254 sufficient for your needs?  If so I will go ahead and close this PR.

doubleyou:
Yeah, I think we're good. We'll just set ERL_FLAGS for -noshell and -noinput. Thanks!

jaredmorrow:
Thanks @doubleyou

PR #266, opened 2013-01-25
Add libstdc++ library to LD_PRELOAD path to find proper symbols

jaredmorrow:
In testing we found a bug caused by loading and incompatible
C++ libary, this change forces the proper libstdc++ libary
to be used on older SmartOS machines.


chardan:
+1

PR #269, opened 2013-01-31
More concise RPM installation notes

Line-Noise:
You can resolve and install RPM dependencies in one step by using yum to install the RPM file. It will automatically download and install anything the RPM depends on.


jaredmorrow:
+1, thanks for the help!

Issue #274, opened 2013-02-14
Riak fails to start on single CPU machines


jaredmorrow:

tylerhannan:
Reproduced with 1.3rc3 both with and without the flag in the vm.args.

Without - won't start on single core
With - starts happily.

+1

PR #275, opened 2013-02-14
Change the erlang vm flags to force SMP use

jaredmorrow:
This addresses basho/riak#274 where Riak
failed to start and run on single CPU machines.


jaredmorrow:
+1 by proxy  attached to the bug itself

PR #283, opened 2013-02-25
We insist on a minimum of 5 nodes in a cluster,

joedevivo:
let's start developers off right with 5 dev nodes and get them in the habit of 5 nodes early


alexmoore:
+1.  Docs/tutorials will have to be updated.

jonmeredith:
+1

PR #284, opened 2013-02-28
Move riak-admin script reference from core to KV

engelsanchez:


engelsanchez:
This can not be merged until basho/riak_kv#497 has been merged in first

jrwest:
+1

engelsanchez:
Dropping for now as there are many other places doing the same (mentioning riak-admin) in core. We need a better solution.

Issue #278, opened 2013-02-21
Consider moving `ulimit` checks to nodetool


jaredmorrow:
Right now we check `ulimit -n` in the `riak` script itself which tells you the ulimit on the active shell process, not necessarily the Erlang VM.  For users who give Erlang special limits the shell can falsely warn that ulimit is set too low.

evanmcc:
If we're considering this, we should also consider moving the check into
riak_kv and logging from there. Maybe a start-up check and log of all the
system's arbitrary limits?

jaredmorrow:
@evanmcc also a great idea

savar:
yep.. would be great..

evanmcc:
These two commits should take care of the issue.  Any opinions?

jaredmorrow:
Still no riak_kv reviewer??  @evanmcc crack the whip!

evanmcc:
jwest is on the case

evanmcc:
both PRs merged, closing

jaredmorrow:
Awesome, thanks @evanmcc

PR #286, opened 2013-03-04
move and clarify ulimit check

evanmcc:
partial fix for #278


jaredmorrow:
This commit is fine with me, +1.

evanmcc:
Although this should likely be reverted or altered if we don't go with the other commit on #278

PR #292, opened 2013-03-15
riak-admin reformat-indices

jrwest:
`riak-admin` support for basho/riak_kv#504. targeted at 1.3 branch.

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -604,12 +604,22 @@ case "$1" in
         shift
         $NODETOOL rpc riak_core_console transfer_limit "$@"
         ;;
+    reformat[_-]indexes)
+        if [ $# -gt 3 ]; then
+            echo "Usage: $SCRIPT $1"
+            echo "       $SCRIPT $1 --downgrade"
+            echo "       $SCRIPT $1 <concurrency> [--downgrade]"
+        fi
+        ensure_node_running
+        shift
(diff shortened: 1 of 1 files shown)


rel/files/riak-admin
```diff
@@ -604,12 +604,22 @@ case "$1" in
         shift
         $NODETOOL rpc riak_core_console transfer_limit "$@"
         ;;
+    reformat[_-]indices)
```
russelldb: indices or indexes? I prefer the former, but a long time ago I think Basho made a choice to stick with the latter.

jrwest: i lean towards the former as well mostly because of `riak_core_ring` (e.g. https://github.com/basho/riak_core/blob/master/src/riak_core_ring.erl#L42)

jj1bdx: So far I only see "indexes" too, though I personally prefer the former. No objection on sticking to "indexes" either.

jrwest: but if we decided on "indexes" a while ago will change. realizing `riak_core_ring` is totally different kind of indices.

russelldb:
Tested on some legacy indexes, worked as advertised. +1

jaredmorrow:
Friendly reminder to add a patch to riak_ee as well.   'git format-patch' and 'git am' is how I usually accomplish it easily so you don't need to do another redundant PR.

jrwest:
changed indices -> indexes

PR #293, opened 2013-03-16
Update rebar.config to point basho/yokozuna.git

jj1bdx:
- Since the repository is moved under basho org,
  dependency point on rebar.config must be fixed
  to properly reflect the changes.


jaredmorrow:
+1

Issue #288, opened 2013-03-10
embedded option 


dergraf:
In https://github.com/basho/riak/blob/master/rel/files/riak#L273 the erl option -embedded is used. The option -embedded doesn't exist in http://www.erlang.org/doc/man/erl.html, but the option '-mode embedded' therefore exists. If you use '-mode embedded'  instead a staging release can't boot anymore since it can't resolve the symlinked dependencies (missing vsn-suffix). 
My guess is that booting a staging release worked because of 'erl' ignoring the' -embedded' option and switched back to the default interactive mode. 

I tested this behavior on Mac OSX 10.8.2 and Erlang R15B03 (erts-5.9.3.1)

jaredmorrow:
Thanks for reporting this, looks like it is an ancient holdover from Erlang R13 days.

jaredmorrow:
Fixed in the above PR's, thanks for reporting it!

PR #310, opened 2013-04-12
Remove incorrect `-embedded` flag from riak startup command

jaredmorrow:
Fixes basho/riak#288


Vagabond:
+1, as this option is having no effect.

I still wonder if we shouldn't try to fix embedded mode, but we can fight about that later.

PR #314, opened 2013-04-17
Move SSL distribution items into a regular configuration file.

seancribbs:
Before this change, `sed` was used to find some "magic comments" in `etc/vm.args` and slurp them into a file that is writeable by the `riak` script, so that `nodetool` could use that file as arguments to its init process, and thus be able to connect to a Riak node using the SSL distribution.

This is wholly unnecessary, given that the emulator can accept multiple `-args_file` command-line arguments. Instead, we now separate the totally-optional-and-mostly-unused SSL settings into `etc/ssl_distribution.args`, which can be used independently by `nodetool` and in conjunction with `etc/vm.args` by the normal node boot process.

In order to support upgrades, the `riak` script checks for the presence of the old `sed`-slurped file. If it is present and contains active configuration items, a warning is presented and the script exits non-zero. If the file exists and only contains comments, it is deleted. Downgrades are not explicitly supported.

### DIFF
--- a/rel/files/nodetool
+++ b/rel/files/nodetool
@@ -1,5 +1,5 @@
 #!/usr/bin/env escript
-%%! -args_file {{platform_data_dir}}/ssl_distribution.args_file
+%%! -args_file {{platform_etc_dir}}/ssl_distribution.args
 %% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
 %% ex: ft=erlang ts=4 sw=4 et
 %% -------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


seancribbs:
Suggested reviewers: @jaredmorrow @slfritchie @Vagabond

Vagabond:
Looks legit at first glance.

jaredmorrow:
I need to look at how to do this right in node_package because the Riak script and nodetool are about to disappear. 

I don't think it will be a problem.

jaredmorrow:
After looking it over, I don't see anything wrong with it in regards to being generic enough to go into node_package.

seancribbs:
@jaredmorrow I'll look into porting it to node_package.

slfritchie:
Ping ... is this PR still relevant?

seancribbs:
@slfritchie Nope!

PR #324, opened 2013-05-10
Added yokozuna enable option to app.config

ksauzz:
I added yokozuna enabled option to app.config. thanks.

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -320,6 +320,7 @@
                 ]},
 
 {yokozuna, [
+            {enabled, false},
             {solr_port, "{{yz_solr_port}}"},
             {solr_jmx_port, "{{yz_solr_jmx_port}}"},
             {solr_vm_args,


jj1bdx:
@ksauzz <del>Yokozuna will NOT be enable WITHOUT this app.config modification, right?</del>
To enable yokozuna, you need to set `{enabled, true}`. Correct me if I'm wrong.

jj1bdx:
@ksauzz So this modification adds the flag for enabling the yokozuna functionality, but the flag itself is set to `false`, to prevent yokozuna to be accidentally enabled. Correct me if I'm wrong.

ksauzz:
@jj1bdx Yes. Changing the flag to `true` is required to run yokozuna. This modification is to just show riak user the existence of the option. Unfortunately, there is no method to know how to run yokozuna except reading source code now. So I thinks the option should be revealed on the configuration file like riak_control. Also we should update 'Getting Started' on README of yokozuna after the merging.

jj1bdx:
@ksauzz: +1. This patch is also applicable to the branch `yz-merge-master` (with 1 fuzz). A brief test here on OSX 10.8.3 with R15B03-1, the `enabled` flag worked as intended when changing the settings between `true` and `false`.
(BTW comment on wording: reveal -> disclose)

@rzezeski - Could you take a look at this?

rzezeski:
@ksauzz Yes, the enabled option should be in the Yokozuna README.  This last week has been a bit hectic for me getting ready for RICON.

@jj1bdx Don't worry about `yz-merge-master` for now.  Once 1.4 drops all these side branches will go away as proper PRs are made to merge Yokozuna into Riak/KV master.

ksauzz:
Thanks. ok, I'll try to update that.

Issue #309, opened 2013-04-11
Raise ETS table limit substantially


slfritchie:
ETS tables are cheap: we should be packaging up vm.args with much bigger table limits.  99 times out of 50, customers are bitten by too few ETS tables, not too few bytes to store all of the data in ETS tables.  (Except for memory backend abuse, of course.)

slfritchie:
Closing: 327 is done

PR #327, opened 2013-05-21
Increase ERL_MAX_PORTS and ERL_MAX_ETS_TABLES

slfritchie:
Addresses #309.

Memory effect on my OS X laptop: starting from scratch with empty
data directories, process VSZ increases about 33MB.


evanmcc:
+1, I have a nearly identical PR that got sidetracked last week by customer issues.

Issue #317, opened 2013-04-24
Riaknostic issue when compiling rel or devrel from source


glickbot:
Haven't had time to dig deeper yet, but when compiling via 'make rel' or 'make devrel', the command "$RIAK/bin/riak-admin diag" results in the following error:
$ riak/bin/riak-admin diag
escript: exception error: undefined function riaknostic:main/1
  in function  escript:run/2 (escript.erl, line 741)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_it/1
  in call from init:start_em/1

( Mac OS, riak-1.3.1 source tarball )

I'll add more info when I get a chance to investigate further

jaredmorrow:
Fixed in PR to change riaknostic from an escript to an app.

PR #329, opened 2013-05-24
Riaknostic no longer escript. Alter build process accordingly.

bsparrow435:
Clean up rebar.config, add riaknostic to reltool, alter riak-admin diag to accommodate non-escript behavior.

Also fixes: basho/riak#317

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,22 +48,10 @@ test: deps compile testclean
 ##
 ## Release targets
 ##
-rel: deps compile generate riaknostic-rel
+rel: deps compile generate
 
 relclean:
 	rm -rf rel/riak
-
-##
-## Riaknostic targets
-##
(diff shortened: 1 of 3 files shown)


jaredmorrow:
Things I checked:
- Devrel build & riak-admin diag testing
- dist build & source tarball build while offline

Two things need to get done if possible, but not before this PR gets in.  
- The riak_test riaknostic test needs to be changed for the new behavior
- The readme for riaknostic needs to be updated for the new behavior

@bsparrow435 I removed extra riaknostic calls from the Makefile that weren't needed anymore.

+1

bsparrow435:
Merging, opened issues against riaknostic for README and riak_test for riaknostic_rt.erl changes. Merging.

PR #331, opened 2013-05-24
Lower net_ticktime to check for aliveness more often

jaredmorrow:


slfritchie:
I've updated both repos of "jem-net-ticktime" with new words.  If you find them inoffensive, +1 to merge.  If offensive, edit them yourself and then don't wait for more review: ship it.

PR #333, opened 2013-05-28
Add health check deprecation note to config

engelsanchez:
/cc @jaredmorrow @jtuple 
Let me know if you have any comments on the wording of this note and stuff. Should be moved to riak_ee and 1.4 before release.

Health checks will be disabled anyway, but this should clear some confusion on users of fresh deployments

### DIFF
--- a/rel/files/app.config
+++ b/rel/files/app.config
@@ -65,7 +65,12 @@
               %% If disabled, health checks registered by an application will
               %% be ignored. NOTE: this option cannot be changed at runtime.
               %% To re-enable, the setting must be changed and the node restarted.
-              {enable_health_checks, true},
+              %% NOTE: As of Riak 1.3.2, health checks are deprecated as they
+              %% may interfere with the new overload protection mechanisms.
+              %% If there is a good reason to re-enable them, you must uncomment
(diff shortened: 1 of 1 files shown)


jaredmorrow:
I like the wording.  If someone upgrades from 1.3.1 -> 1.3.2, and they have health checks turned on, what happens?

jtuple:
Health checks are forced disabled in `1.3.2` unless people explicitly enable them by uncommenting the line Engel provides here in `app.config`. This config option is entirely new in `1.3.2`. There are other health check flags that exist in earlier releases, but without this new option provided, KV health checks won't be enabled. See [this commit](https://github.com/basho/riak_kv/commit/5860d68a8ac7eed3eb33e9346d90b946864a488f) for details.

EDIT: Rather, the `riak_kv` option Engel mentions in his `app.config` comment is the new option in `1.3.2`. Without **also** adding this option, KV health checks are disabled regardless of the `riak_core` setting.

jaredmorrow:
With that said +1 from me @engelsanchez 

Please 'git format-patch' and 'git am' into riak_ee.  If you would like just ping me and I can merge 1.3->master as well.  I've already done it once and my system should remember the conflict resolutions.

PR #336, opened 2013-06-07
update default configs for 1.3.2

evanmcc:
- add +P to vm.args and comment out +scl +zdss by default
- remove health check stuff from app.config
- add fsm_limit to app.config


jaredmorrow:
+1, lets do this

PR #342, opened 2013-06-25
riak-admin downgrade-objects

jrwest:
wire up riak_kv_console:reformat_objects/1 to riak-admin. command
is called "downgrade-objects" to avoid confusion since that is the
only thing currently supported

relies on: https://github.com/basho/riak_kv/pull/585

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -24,10 +24,11 @@ SCRIPT=`basename $0`
 usage() {
     echo "Usage: $SCRIPT { cluster | join | leave | backup | restore | test | "
     echo "                    reip | js-reload | erl-reload | wait-for-service | "
-    echo "                    ringready | transfers | force-remove | down | "
+    echo "                    ringready | transfers | force-remove | down |"
     echo "                    cluster-info | member-status | ring-status | vnode-status |"
     echo "                    aae-status | diag | status | transfer-limit | reformat-indexes |"
-    echo "                    top [-interval N] [-sort reductions|memory|msg_q] [-lines N] }"
+    echo "                    top [-interval N] [-sort reductions|memory|msg_q] [-lines N] |"
+    echo "                    downgrade-objects }"
 }
 
 cluster_admin()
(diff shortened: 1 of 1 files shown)


rel/files/riak-admin
```diff
@@ -509,6 +509,15 @@ case "$1" in
         shift
         $NODETOOL rpc riak_kv_console reformat_indexes "$@"
         ;;
+    downgrade[_-]objects)
```
evanmcc, line 513: This option is not documented in generic usage for the script at the top.

rel/files/riak-admin
```diff
@@ -509,6 +509,15 @@ case "$1" in
...
+        if [ $# -lt 2 ]; then
```
evanmcc, line 514: this isn't catching the too many arguments case.  this may or may not be an issue, but the error message coming from riak isn't terribly specific.

rel/files/riak-admin
```diff
@@ -509,6 +509,15 @@ case "$1" in
...
+            echo "Usage: $SCRIPT $1 <kill-handoffs> [<concurrency>]"
+            exit
+        fi
+        node_up_check
+        shift
+        $NODETOOL rpc riak_kv_console reformat_objects "$@"
```
evanmcc, line 520: I'd really like a message here letting the user know that a background reformat has been launched and to check the logs for completion status.

rel/files/riak-admin
jrwest, line 513: added in https://github.com/basho/riak/commit/e6f4b3f98550eefc2eb01831ee8192c2b4c3c07c

rel/files/riak-admin
jrwest, line 514: handled in `riak_kv`

rel/files/riak-admin
jrwest, line 520: handled in `riak_kv`

evanmcc:
:+1:, looks good.

PR #346, opened 2013-06-28
Added support for managing index repairs through riak-admin

cdahlqvist:
This PR depends on the following riak_kv PR: https://github.com/basho/riak_kv/pull/591

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -12,7 +12,7 @@
 {deps, [
        {lager_syslog, "1.2.2", {git, "git://github.com/basho/lager_syslog", {tag, "1.2.2"}}},
        {cluster_info, "1.2.3", {git, "git://github.com/basho/cluster_info", {tag, "1.2.3"}}},
-       {riak_kv, "1.3.2", {git, "git://github.com/basho/riak_kv", {tag, "1.3.2"}}},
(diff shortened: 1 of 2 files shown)


jaredmorrow:
Please remove the rebar.config file from the PR as it would change the rebar.config to point to a specific branch of riak_kv.

cdahlqvist:
Rolled back my change to rebar.config to original 1.3.2 value.

slfritchie:
Ping ... is this PR needed for Riak 2.0?  The prerequisite PR is still open and might be close without action because of other 2.0 features.  Thoughts?

jaredmorrow:
2i Index repairs were added in https://github.com/basho/riak/pull/459, so I'm going to close this considering the riak_kv PR is also not merged.  We can resurrect if the riak_kv PR ends up being adjusted and put in.

jrwest:
Just to clarify, the repair mentioned here is the existing repair mechanism [1] before the work done by @engelsanchez for 1.4.4+ and 2.0. Since we have a new mechanism and we never decided if indeed we wanted to expose the original method used here, I think closing is fine.

[1] http://docs.basho.com/riak/latest/ops/running/recovery/repairing-indexes/

engelsanchez:
Wow, I got all sorts of confused when I got this today. We really have mega-overload problems on the word index :)

jaredmorrow:
Yes, sorry too many "index"s in the code

PR #350, opened 2013-07-10
Make riak compatible with R16

evanmcc:
This PR depends on the following PRs:
- https://github.com/basho/riak_search/pull/138
- https://github.com/basho/riak_kv/pull/602
- https://github.com/basho/webmachine/pull/156
- https://github.com/basho/riak_core/pull/348
- https://github.com/basho/eper/pull/6

Note that some of the changes needed aren't well-covered by unit tests, so if you can think of some way to exercise them, please let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ OVERLAY_VARS    ?=
 
 $(if $(ERLANG_BIN),,$(warning "Warning: No Erlang found in your path, this will probably not work"))
 
+# The code below figures out the OTP release and introduces a macro at 
+# build and test time to tell later released to use the new hash
+# functions introduced in R15B02.  Older versions still use the old
+# hash functions.
 VSN := $(shell erl -eval 'io:format("~s~n", [erlang:system_info(otp_release)]), init:stop().' | grep 'R' | sed -e 's,R\(..\)B.*,\1,')
 NEW_HASH := $(shell expr $(VSN) \>= 16)
 ifeq ($(NEW_HASH),1)
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,18 @@ OVERLAY_VARS    ?=
 
 $(if $(ERLANG_BIN),,$(warning "Warning: No Erlang found in your path, this will probably not work"))
 
+VSN := $(shell erl -eval 'io:format("~s~n", [erlang:system_info(otp_release)]), init:stop().' | grep 'R' | sed -e 's,R\(..\)B.*,\1,')
+NEW_HASH := $(shell expr $(VSN) \>= 16)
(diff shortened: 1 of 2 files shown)


seancribbs:
According to some of @cmeiklejohn's investigation today, the `crypto:hash(Algo, Term)` form emerged in R15B02. Can we really claim compatibility with releases earlier than that in rebar.config?

seancribbs:
Disregard, you've added a macro definition to the build process.

Makefile
```diff
@@ -8,12 +8,18 @@ OVERLAY_VARS    ?=
 
 $(if $(ERLANG_BIN),,$(warning "Warning: No Erlang found in your path, this will probably not work"))
 
+VSN := $(shell erl -eval 'io:format("~s~n", [erlang:system_info(otp_release)]), init:stop().' | grep 'R' | sed -e 's,R\(..\)B.*,\1,')
```
jaredmorrow: It would be nice to have a comment block above this section.  It is not clear what the goal is for all this mayhem outside the context of the PR.

rebar.config
```diff
@@ -1,6 +1,6 @@
 {sub_dirs, ["rel"]}.
 
-{require_otp_vsn, "R14B0[234]|R15"}.
+{require_otp_vsn, "R14B0[234]|R15|R16"}.
```
joedevivo, line 3: Is there a reason to keep R14 compatibility?

evanmcc, line 3: Not knowing whether or not it works on R14, I didn't see any reason to
remove it, given its state of benign neglect over the course of 1.2, 1.3,
and 1.4.  That said, I don't mind removing it, as a signal that we're only
targeting the latest releases of erlang.

cmeiklejohn, line 3: :+1: to being explicit and remove it.

joedevivo, line 3: Do we want to take it as far as saying "A version of Riak is targeted at a particular OTP version" and explicitly require that version? In this case, R16B01.

evanmcc, line 3: My feeling is that we should have all of the versions in there that we're at least willing to take bug reports from in the open source community.  So I think that we should be liberal here and broadly allow R15 and R16, deprecating R15 when we add R17 support sometime next year.

jaredmorrow, line 3: As both a company who is trying to make money and as good stewards to the OSS community I think keeping as much compatibility as possible is a good goal to have.  If R14 is broken with new code and it is hard for us to maintain going forward, then I'm fine to remove it.  If it is just removing it for the sake of cleanup, then I don't feel that's a great reason.  If someone is hitting this line of code, that means they are building from source and already outside of what we claim to support in a production sense, so I don't think that is an issue.

PR #351, opened 2013-07-11
Japanese translation of Release note.

kuenishi:
@ksauzz and @shino translated (reviewed twice!) the 1.4 releasenote. Could someone tell if it is okay to put this translation in this directory?
Or maybe is it better to put this _before_ the release?


jaredmorrow:
I'm +1 on this going where you put it. 

 I also would be +1 if you put it in the top-level right next to the other RELEASE-NOTES file.  LIke other releases I can then move it to the releasenotes directory when a new set of RELEASE-NOTES comes out.  I don't think you need to hide the Japanese translation away in another directory.  I will let you make the call.

Thanks for doing the work!

kuenishi:
Moved to toplevel. If there's more suitable filename, fix it please.

PR #360, opened 2013-08-08
Add basic performance smoke test

engelsanchez:
This is a simple tool meant to be a sanity check for whether a change to
Riak causes a severe performance degradation in the main KV code paths.

First try at this little sucker. /cc @jtuple @evanmcc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ perfdev : all
 	mkdir -p perfdev
 	(cd rel && ../rebar generate target_dir=../perfdev overlay_vars=vars/perf_vars.config)
 
-perf: perfdev
+perf:
 	perfdev/bin/riak stop || : 
 	perfdev/bin/riak start
 	perfdev/bin/riak-admin wait-for-service riak_kv 'perfdev@127.0.0.1'
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,19 @@ dev% : all
 	rel/gen_dev $@ rel/vars/dev_vars.config.src rel/vars/$@_vars.config
 	(cd rel && ../rebar generate target_dir=../dev/$@ overlay_vars=vars/$@_vars.config)
 
+perfdev : all
+	perfdev/bin/riak stop || :
+	rm -rf perfdev
+	mkdir -p perfdev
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -78,6 +78,19 @@ dev% : all
 	rel/gen_dev $@ rel/vars/dev_vars.config.src rel/vars/$@_vars.config
 	(cd rel && ../rebar generate target_dir=../dev/$@ overlay_vars=vars/$@_vars.config)
 
+perfdev : all
+	perfdev/bin/riak stop || :
+	rm -rf perfdev
+	mkdir -p perfdev
+	(cd rel && ../rebar generate target_dir=../perfdev overlay_vars=vars/perf_vars.config)
+
+perf: perfdev
```
evanmcc: I'd like the actual perf stuff to be a separate target, `make runperf` or something.  This will allow easier tweaking when you don't need to rebuild then entire perfdir.

Makefile
```diff
@@ -78,6 +78,19 @@ dev% : all
 	rel/gen_dev $@ rel/vars/dev_vars.config.src rel/vars/$@_vars.config
 	(cd rel && ../rebar generate target_dir=../dev/$@ overlay_vars=vars/$@_vars.config)
 
+perfdev : all
+	perfdev/bin/riak stop || :
+	rm -rf perfdev
+	mkdir -p perfdev
+	(cd rel && ../rebar generate target_dir=../perfdev overlay_vars=vars/perf_vars.config)
```
evanmcc, line 85: Also it'd be awesome here if instead of the rel path we altered the stagedevrel path, then someone could make a code change, run `make compile` and then `make runperf` without having to redo any tweaks to the devrel (after it is removed and rebuilt).

evanmcc, line 85: I addressed this in my commits.

evanmcc:
Updated, for when you have a second to look at it.   +1, although we may want someone else to glance at it now that we've both got commits here.

evanmcc:
Also I guess that this should be moved to develop, rather than merged to master.

evanmcc:
Retargeted PR has merged, closing.

PR #374, opened 2013-09-05
fixed trailing dots in schema

joedevivo:
trailing dots no longer satisfy that function call.


engelsanchez:
+1 Without this, my riak tests were failing, now pass. Ship it! :dancer: :boat:

PR #375, opened 2013-09-06
Add Yokozuna

rzezeski:
Update various files to include Yokozuna.


coderoshi:
Should note that this first requires https://github.com/basho/yokozuna/pull/166 because of yokozuna.schema

rzezeski:
Addresses basho/yokozuna#155

PR #385, opened 2013-09-14
Update riak-admin wait-for-service to validate node name

deanproctor:
Updates riak-admin wait-for-service to check if the TARGETNODE is known to the local node before checking services.  Addresses https://github.com/basho/riak_core/issues/151

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -338,6 +338,15 @@ case "$1" in
                 continue
             fi
 
+            # Make sure the requested node is known before checking service status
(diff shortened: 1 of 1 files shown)


evanmcc:
I applaud the intention here, but I wonder if this isn't best addressed in core itself, if the node-watcher shouldn't bail if it doesn't know about the node.  One thing that could be addressed on this side would be to make if so that if `TARGETNODE` is blank to just use the local node, which I think would address most instances of this problem more directly.

PR #387, opened 2013-09-18
do not make _yz_default data directory

beerriot:
this directory and its files are not needed as of basho/yokozuna#171 (final cleanup for basho/yokozuna#143)


coderoshi:
+1

PR #389, opened 2013-09-19
Tell reltool to exclude some empty folders

Vagabond:
Empty folders make some package managers mad, like FreeBSD.


jaredmorrow:
Tested on OSX 'make rel' and the only empty directories I still see are:

```
./data/ring
./lib/yokozuna-v0.9.0-17-gff0cacb/priv/solr/solr-webapp
./log
```

Which are needed.

+1

PR #390, opened 2013-09-20
Make Yokozuna port numbers consistent with EE

rzezeski:
When creating Yokozuna patch for EE I realized that the 5 & 6 suffixes were already taken by other services.  Therefore I switched Yokozuna to use 3 & 4 so it could be consistent across both packages.


jaredmorrow:
+1

PR #392, opened 2013-09-25
Add missing rel storage value

engelsanchez:
With the addition of the smoke test, storage backend was made into a
variable and perf_dev and devrel targets were updated, but not the rel
target. This adds the missing default storage of "bitcask".


seancribbs:
:+1:

Vagabond:
+1

PR #393, opened 2013-09-25
Move the stub 'riak' app into a subdir to avoid confusing rebar

Vagabond:
Right now, if you try to do ./rebar eunit apps=riak_kb from the toplevel
of riak, rebar will just return. This is because it sees the ebin
directory at the toplevel and somehow it gets confused. The more
'release' friendly directory path is to have apps/riak/(ebin|src).

This branch makes that change and updates the rebar.config and
reltool.config files accordingly

A cuttlefish fix is also needed: basho/cuttlefish#35

cc @jaredmorrow @slfritchie


jaredmorrow:
+1

Make sure to patch riak_ee

PR #394, opened 2013-09-26
Make the node parameter optional for `riak-admin wait-for-service`

deanproctor:
This changes the behavior of `riak-admin wait-for-service` to default to the local node if no target node is supplied.  I believe this behavior is more intuitive for new users.


jaredmorrow:
+1

jaredmorrow:
I'll patch to riak_ee

PR #396, opened 2013-10-02
fixed handing riak_control.user

ksauzz:
Generated app.config doesn't include riak_control username and pass as userlist even if it's configured.

before fix

```
% grep riak_control app.2013.10.02.02.33.17.config
 {riak_control,[{enabled,true},{auth,userlist},{userlist,[]},{admin,true}]},
```

after fix

```
% grep -A 4 riak_control app.2013.10.02.10.09.40.config
 {riak_control,
     [{enabled,false},
      {auth,userlist},
      {userlist,[{"user","pass"}]},
      {admin,true}]},
```

### DIFF
--- a/rel/files/riak.schema
+++ b/rel/files/riak.schema
@@ -687,15 +687,12 @@ end}.
 {translation,
 "riak_control.userlist",
 fun(Conf) ->
-  UserList1 = lists:filter(
+  UserList = lists:filter(
     fun({K, _V}) -> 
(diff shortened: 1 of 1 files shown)


seancribbs:
:+1: Thanks for the fix!

jaredmorrow:
This also needs to be patched into riak-ee, you can attempt 'git format-patch' on this commit and then 'git am' in the riak_ee repo.

ksauzz:
Thanks. I'll also apply the changes to riak-ee.

PR #397, opened 2013-10-02
Add warning/max object limits

engelsanchez:


evanmcc:
I think that these sizes are OK, but fwiw mostly the problematic sibling counts were much lower than this when I was a CSE.  warning at 25 and stopping at 100 sounds more reasonable to me, but I might be over-conservative.

engelsanchez:
The sizes were pulled out of a hat. I would feel more comfortable with anything recommended from CSE experience.

evanmcc:
K. Drop them to 25/100 and :+1:, we can tune more with CSE input post-ricon.

engelsanchez:
@evanmcc I've changed the limits based on your CSE experience recommendation and I've opened a PR to apply a copy of this to Riak EE here https://github.com/basho/riak_ee/pull/182

PR #401, opened 2013-10-02
Remove bundled SSL certificates.

cmeiklejohn:
@jaredmorrow Remove the bundled SSL certificates.  I'll have a separate PR for Riak EE.


Vagabond:
Also +1

PR #403, opened 2013-10-02
Remove modules.

cmeiklejohn:
These are not used by Riak Control at all, and won't be in the
foreceeable future.


cmeiklejohn:
Please let me know when this is merged, so I can open a PR against enterprise.

joedevivo:
+1

PR #405, opened 2013-10-03
Fix a schema comment to reflect the proper accepted value.

jgnewman:
- The accepted value should be `off` rather than `false`.

### DIFF
--- a/rel/files/riak.schema
+++ b/rel/files/riak.schema
@@ -642,7 +642,7 @@ end}.
 ]}.
 
 %% riak_control config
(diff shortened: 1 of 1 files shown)


joedevivo:
+1

cmeiklejohn:
Was this also fixed for riak_ee?

jgnewman:
My bad, didn't realize I had to change it in 2 places.  Doing it now.

[riak_ee PR #179](https://github.com/basho/riak_ee/pull/179)

PR #406, opened 2013-10-03
move leveldb cuttlefish stuff to eleveldb/priv/eleveldb.schema

matthewvon:
This includes leveldb options for mv-flexcache ... so it expects eleveldb's mv-flexcache and leveldb's mv-flexcache


joedevivo:
+1

PR #419, opened 2013-10-22
Resolve reip so that it uses the proper cuttlefish configuration files.

cmeiklejohn:
As best as I can tell, we don't need to worry about the configuration
being edited, because it appears that reip only modifies the ring files,
and not the vm.args or app.config.


joedevivo:
tested +1

PR #420, opened 2013-10-23
Add security enable/disable/status commands

Vagabond:
See also basho/riak_core#437


engelsanchez:
:+1: More notes in the main PR https://github.com/basho/riak_core/pull/437

PR #421, opened 2013-10-23
remove bucket-type reset from riak-admin

jrwest:
its semantics right now are a bit unclear so lets not make it user-facing. We'll revisit soon.
- [x] apply to riak_ee https://github.com/basho/riak_ee/commit/931752d44923e2f1d7f5cd8cb03ea822a951d44f


Vagabond:
+1

PR #422, opened 2013-10-25
Add riak_ensemble to reltool.config

jtuple:
Add [riak_ensemble](http://github.com/basho/riak_ensemble) as an included/loaded (but not started) app. `riak_ensemble` is actually started by `riak_core` via basho/riak_core#441

### DIFF
--- a/rel/reltool.config
+++ b/rel/reltool.config
@@ -29,7 +29,8 @@
          riak_control,
          erlydtl,
          riak_auth_mods,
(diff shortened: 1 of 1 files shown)


jaredmorrow:
+1 whenever riak_core PR merges and brings it in

jtuple:
Merging as part of basho/riak_kv#710 and basho/riak_core#441

PR #425, opened 2013-10-26
Add reset to riak-admin bucket-type

lafka:
Reset is implemented in riak_kv_console but not in riak-admin


jrwest:
@lafka `reset` was purposefully removed from `riak-admin` temporarily [1] because it is not ready to be exposed to users and has ambiguous results. It will be fixed for the final release of 2.0.

[1] https://github.com/basho/riak/pull/421

PR #431, opened 2013-10-31
Correct graphviz node names with hyphen

reiddraper:
Fix comes from a suggestion from @shino, in basho/riak_cs#711. If accepted, I'll add to EE.


jaredmorrow:
+1, worksforme

PR #435, opened 2013-11-12
preparation for configurable name of cuttlefish conf files

joedevivo:
In order to make cuttlefish accept a conf filename as a configuration parameter, I needed to add a default which is not `riak.conf`. it's `cuttlefish.conf`. Anyway, if this lands in riak first, it won't be an issue.

Also, I figured out that I didn't need the cuttlefish plugin in the main rebar.config, only the one in rel.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -25,4 +25,4 @@
         {rebar_lock_deps_plugin, ".*", {git, "git://github.com/seth/rebar_lock_deps_plugin.git", {branch, "master"}}}
        ]}.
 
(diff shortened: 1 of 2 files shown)


seancribbs:
So, I get the reasoning for the config, but why remove the plugin?

joedevivo:
unnecessary. only the plugin in `rel/rebar.config` was ever being productive.

seancribbs:
Ok then, :+1:

Issue #439, opened 2013-11-14
Solaris source builds will not start


jaredmorrow:
### Bug description

When making a release with `make rel` on Solaris 10, the `riak` script will not run properly.  Failing with a `bad substitution` error.  This failure does not happen in the packages, just from source.

The problem shows up like:

```
build-solaris-10u9-64.(~/dev/riak/rel/riak) ./bin/riak help
./bin/riak: bad substitution
```
### Why this is happening

This is due to the `runner_script_dir` containing a POSIX bourne shell substitution that Solaris's non-posix /bin/sh supports.  You can see [this problem in the vars.config](https://github.com/basho/riak/blob/c99a53f4e90742b37e558097de867a38a4617c11/rel/vars.config#L46)

This gets substituted [here](https://github.com/basho/node_package/blob/ba81fc9bdbd70f5a5f4091c8e72b1aaf21cdf376/priv/base/runner#L8) before we get a chance to reload into to /bin/ksh on Solaris [in the env.sh file here](https://github.com/basho/node_package/blob/ba81fc9bdbd70f5a5f4091c8e72b1aaf21cdf376/priv/base/env.sh#L14)
### Why testing did not find this earlier

This problem did not get found in testing due to testing being done with the packages on Solaris.  The packages use absolute paths defined in the packaging templates.  For example, Solaris defines its directories [in the solaris.template file in node_package](https://github.com/basho/node_package/blob/ba81fc9bdbd70f5a5f4091c8e72b1aaf21cdf376/priv/templates/solaris/solaris.template#L14)
### Solution

Replace substitution with Solaris /bin/sh code (hacks) that will work on all platforms.

jaredmorrow:
Fixed in above PR's.

PR #440, opened 2013-11-14
Fix path substitution code to fix Solaris source builds

jaredmorrow:
Addresses: basho/riak#439


jaredmorrow:
Merging based on PR +1 in riak_ee https://github.com/basho/riak_ee/pull/193#issuecomment-28531959

PR #443, opened 2013-11-19
remove ebloom app from rel/reltool.config

vinoski:
The ebloom app dependency was removed from the riak_kv repo in commit
8bd8cde, and is not used in riak, so remove it from reltool.config.


cmeiklejohn:
release build; riak started.  :+1:

PR #456, opened 2013-12-18
Riak debug additions

slfritchie:
- Add 'blockdev' ouput to the platform (Linux) that supports it
- Add cluster-info output to the Riak commands section

### DIFF
--- a/rel/files/riak-debug
+++ b/rel/files/riak-debug
@@ -407,6 +407,12 @@ if [ 1 -eq $get_riakcmds ]; then
     dump riak_diag "$riak_bin_dir"/riak-admin diag
     dump riak_repl_status "$riak_bin_dir"/riak-repl status
 
+    CI=`pwd`/cluster-info.html
+    touch $CI
+    chmod 666 $CI
+    dump cluster-info riak-admin cluster-info $CI
+    chmod 444 $CI
+
     # Get the latest ring file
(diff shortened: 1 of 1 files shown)


slfritchie:
Comments @bsparrow435 or @joecaswell or @jmshoffs0812 or other CSEs?

bsparrow435:
I'm a bit wary of cluster-info being run by default as it's pretty heavy handed.. but I have not seen it take down a node before so i'm fine with this change. :+1:

kuenishi:
My understanding is, cluster-info is cluster-wide by default while existing riak-debug is per-node. So I think it'd be better by making it local like:

```
dump cluster-info riak-admin cluster-info $CI local
```

bsparrow435:
Good point @kuenishi, I agree. Otherwise, collecting a `riak-debug` from each node in the cluster will generate a lot of unnecessary load.

slfritchie:
I'm not involved in a lot of support work lately, but when I am, it's annoying to have to go back and ask for diagnostic info batch #2.  If riak debug is supposed to be the be-all, do-all, one-stop shop, then why two stops?

Kota's point, however, I'll concede.  We recently had a customer who ran riak debug on all nodes in the same cluster ... if that was done via Ansible or multi-session tmux, for example, then it could very well be all running simultaneously.  I'll add a local only flag to the cluster info invocation, then merge.  Thanks!

PR #467, opened 2014-01-10
Fixed devrel cuttlefissue. Needs newest cuttlefish to work

joedevivo:
This will fix the issue with devrels introduced this week. Requires basho/cuttlefish#105 to be merged first.


Vagabond:
Make this a basic option, and then +1 from me.

PR #482, opened 2014-02-05
Add a deprecation notice onto search-cmd

coderoshi:
Yokozuna is the new hotness. Don't let people forget it with this classy deprecation notice.

basho/yokozuna#272

### DIFF
--- a/rel/files/search-cmd
+++ b/rel/files/search-cmd
@@ -22,5 +22,8 @@ cd $RUNNER_BASE_DIR
 SCRIPT=`basename $0`
 
 # Make sure the local node IS running
+echo "** This tool is for a deprecated version of Riak Search."
+echo "** This will not work for Yokozuna. If you have yet to, please upgrade."
+echo ""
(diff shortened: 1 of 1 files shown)


jonmeredith:
Yokozuna vs Search2 'branding'.

evanmcc:
Also the warning should be suppressable for people who're using search-cmd in scripts.

jaredmorrow:
`echoerr` can be used from the env.sh https://github.com/basho/node_package/blob/develop/priv/base/env.sh#L50 so as to print on stderr and not affect runner scripts.

Also, placeholder to remember to format-patch / am to riak_ee after this is +1'd.

coderoshi:
After painful deliberation, it looks like we're back to the "Riak Search" name.

rzezeski:
+1, don't forget to update EE

coderoshi:
@rzezeski https://github.com/basho/riak_ee/pull/220

PR #488, opened 2014-02-12
RHS schema subs platform_log_dir, {level, advanced} -> hidden

joedevivo:
This PR is to handle two overall cuttlefish related changes:
#### replacing `{level, advanced}` with `hidden`

`{level, advanced}` was getting confused with the `advanced.config` file. basho/cuttlefish#120 added a synonym for `{level, advanced}` called `hidden` and we migrated the riak.schema to that format.
#### Right Hand Side (RHS) substitutions for platform_*_dir

Riak's overall configuration was inconsistent when it came to relative directories. basho/cuttlefish#116 added the ability to set one configuration setting relative to another. So if you change the "parent" setting, the child (if left to it's default) would also change.

For example, changing `platform_log_dir` will change the default location of both `console.log` and `error.log`

Requires:
- [x] Eng +1
- [x] CSE +1
- [x] ProdMgmt +1

### DIFF
--- a/apps/riak/test/riak_schema_test.erl
+++ b/apps/riak/test/riak_schema_test.erl
@@ -4,7 +4,7 @@
 
 basic_schema_test() ->
     Config = cuttlefish_unit:generate_templated_config(
-               ["../../../rel/files/riak.schema"], [], context()),
+               ["../../../rel/files/riak.schema"], [], context(), predefined_schema()),
 
     cuttlefish_unit:assert_config(Config, "lager.handlers",
                                   [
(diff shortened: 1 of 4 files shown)


angrycub:
CSE +1

gburd:
PM +1

seancribbs:
Eng +1

jaredmorrow:
I know this is already closed, but down the road if we look at this PR it has no explanation at all on what it does.  The "why" is completely missing.  I'd prefer the PR get edited, so we can at least get some context.

joedevivo:
@jaredmorrow what are you talking about? This PR tooooooootally has a description.

jaredmorrow:
Magic!  Thanks.

reiddraper:
Can you work some magic to give the commit a description too ;)?

PR #489, opened 2014-02-13
Treat users and groups as distinct concepts

macintux:
To make it easier to reason about and document the new security behavior, this PR splits roles into users and groups.
- Users (but not groups) can be given authentication information (sources).
- Groups (but not users) can be used to cascade privileges to other groups and users.

Where the code does not or cannot distinguish between users and groups, I continue to refer to roles, but the user interface is constrained to users and groups.

There are 3 projects impacted by this change, all with the same branch name (`jrd-security-role-bifurcation`):
- riak
- riak_core
- riak_test

All should be merged at the same time. I will be filing PRs for each.

/cc @Vagabond

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -216,6 +216,9 @@ security_admin()
         print-users)
             $NODETOOL rpc riak_core_console print_users
             ;;
+        print-groups)
+            $NODETOOL rpc riak_core_console print_groups
+            ;;
         print-sources)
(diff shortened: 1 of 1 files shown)


Vagabond:
+1

jaredmorrow:
Is there a complementary one for basho/riak_ee?

Vagabond:
Oh, we should probably do that (or just copy the damn riak-admin script over).

PR #501, opened 2014-03-04
Tweaks to improve user experience

macintux:
This was developed alongside `jrd-security-output` in `riak_core`, but the changes are entirely independent, so the branches can be merged separately or not at all.
- Indicate the security status after an attempt to change it
- Prevent an ugly RPC error if an admin accidentally supplies extra arguments to `print-user` or `print-group`

/cc @Vagabond


Vagabond:
+1

PR #504, opened 2014-03-07
Add print-grants to riak-admin

macintux:
- `print-grants` takes over for `print-user` and `print-group`, which now are just single-entity versions of `print-users` and `print-groups`
- Use `riak_core_console:print_user/1` and `print_group/1` to illustrate changes after successful adds/alters.

Prerequisite `jrd-print-grants` branch exists under `riak_core`.

/cc @Vagabond


Vagabond:
+1

PR #506, opened 2014-03-10
Japanese translation of update to release note 1.4.8

shino:
Will add the same diff to riak_ee after mege


ksauzz:
Looks nice!  :+1:

PR #513, opened 2014-03-21
No longer need all caps for keywords in grant/revoke commands

macintux:
Tied to a forthcoming PR in `riak_core`.

/cc @Vagabond


lordnull:
Works as advertised. :+1:

PR #516, opened 2014-03-25
Add riak-admin search subcommand

rzezeski:
Add a search subcommand to riak-admin for administrating
Yokozuna.

This is related to basho/yokozuna#349.


jaredmorrow:
+1, works for me when I used the yz and kv PR's.

Please patch riak_ee with the same change.

rzezeski:
@jaredmorrow is that a +1 for all three?

jaredmorrow:
My intention was a +1 for just this PR.  It printed correctly, so if no one else gives you a kv +1 soon, I'll check that code tomorrow and upgrade my +1 to be all-inclusive.

rzezeski:
Okay. FWIW KV is simply an export of 3 functions and Yokozuna is a copy/paste of what KV AAE does.

PR #518, opened 2014-03-25
Increase default +zdbbl setting from 1MB to 32MB

### LINKED ISSUE #445: Really increase the default zdbbl buffer size.
We're seeing clusters tip over for customers and prospects alike, making Riak look very bad and contradicting our operational ease and stability story.

Often the cause of these issues boils down to a flood of busy_dist_port errors in their installations and deployments.

Further confusing is that we put a better value there, but comment it out.  A common convention is to show defaults and comment it out, but that's not the default.  To actually get that value you have to uncomment it.

https://github.com/basho/riak/blob/1.4/rel/files/vm.args#L40

Since the above text says that we at Basho recommend this change anyway, can we just uncomment that line and make the actual default be the default we recommend?  Are there any scenarios where Riak should be deployed or even considered well tested with a zdbbl value of 1024?

jaredmorrow:
Fixes: basho/riak#445

Depends on: https://github.com/basho/cuttlefish/pull/143

This overrides the default found in cuttlefish: https://github.com/basho/cuttlefish/blob/develop/priv/erlang_vm.schema#L178-189
- [x] Eng 
- [x] CSE
- [x] PM

### DIFF
--- a/rel/files/riak.schema
+++ b/rel/files/riak.schema
@@ -168,6 +168,13 @@
   {default, "riak"}
 ]}.
 
+
+%% override zdbbl from 1mb to 32mb
+{mapping, "erlang.distribution_buffer_size", "vm_args.+zdbbl", [
+  {default, "32MB"},
+  merge
+]}.
+
 {{#devrel}}
 %% Because of the 'merge' keyword in the proplist below, the docs and datatype
 %% are pulled from the leveldb schema.


jaredmorrow:
I will format-patch, git am this into basho/riak_ee when it is approved.

joedevivo:
This needs a CSE, PM and ENG +1,

I've already tested it while fixing the underlying cuttlefissue. +1 Eng.

bsparrow435:
+1 CSE

michellep-basho:
+1 from PM

jaredmorrow:
Pushed to riak_ee -> https://github.com/basho/riak_ee/commit/815abc93433abc687c8106835dc975be526d9786

PR #521, opened 2014-03-27
Add switch-to-new-search command

rzezeski:
Add the switch-to-new-search command under the search command. This is
used when migrating from legacy riak_search to yokozuna. It does a
multicall across the cluster handing over HTTP and PB control from
riak_search to yokozuna.

This is needed for issue basho/yokozuna#350.

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -319,6 +319,18 @@ search_admin()
 
             $NODETOOL rpc yz_console aae_status "$@"
             ;;
+        switch[_-]to[_-]new[_-]search)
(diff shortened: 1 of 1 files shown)


coderoshi:
+1 67fb35b
- [x] Manual test

It functions, but one nice-to-have would be a simple message, even just a printout like "Switching interfaces to new search".

jaredmorrow:
Friendly reminder to patch this over to riak_ee after merge.

rzezeski:
Applied to riak_ee as commit `66ba173d72c60186b864b87e35a97e437e7319da`

PR #523, opened 2014-03-27
Reltool.config for new multi_backend.schemas

joedevivo:
@jrwest the final piece of basho/eleveldb#80 and basho/bitcask#157


jrwest:
things build, config works. +1 merge

PR #524, opened 2014-03-27
Add man page docs for new search command

rzezeski:
Add man page docs for the new search command as well as its
sub-commands.

This is for PRs basho/yokozuna#350 and basho/yokozuna#349.

To view the man page run the following:

```
gzip -dc doc/man/man1/riak-admin.1.gz | nroff -man | less
```

PR #541, opened 2014-05-13
Update riak-admin and riak man pages

coderoshi:
The riak-admin and riak man pages were horrible out of date. The options were to update them, or delete them. Clearly, the former was the chosen path.

To test, just run man against the gz files.

```
man ./doc/man/man1/riak-admin.1.gz
man ./doc/man/man1/riak.1.gz
```
#524 opened the door
- [x] Also merge into EE


reiddraper:
@coderoshi I dig the 'Also merge into EE' reminder. We should start adopting that until we've rolled them into the same repo.

kellymclaughlin:
In the section describing the `config` option on the riak page there is a misspelling, _configration_, at the end of the description of the `effective` sub-option.

kellymclaughlin:
Also in the `config` section in the description of the `describe` sub-option: _descrption_

kellymclaughlin:
On the `riak-admin` page under the security sub-commands section, the description of the `ciphers` option uses _ciphers_ as the plural and _cypher_ as the singular. Why not just cipher? Not a big deal, just noticed the difference and was curious.

kellymclaughlin:
Do you think we should we add some sort of disclaimer to the backup and restore command descriptions saying "Do not use! These do not work! DANGER, DANGER!"?

kellymclaughlin:
On the `riak` page the description of `attach` uses existed instead of exited: _...which can be existed more safely_

kellymclaughlin:
:+1:

PR #543, opened 2014-05-16
add 'riak-admin bucket-type reset <type>' command

dave-re:
I have added 'reset' sub command to 'riak-admin bucket-type' command.


jrwest:
the `reset` command will not be supported in Riak 2.0 and was previously removed: https://github.com/basho/riak/pull/421

PR #548, opened 2014-05-22
Update rebar.config to restrict erlang version to R16

kellymclaughlin:
The latest development has been against R16 and the `HEAD` of `develop` no
longer builds with versions less than R16. [Cuttlefish](https://github.com/basho/cuttlefish/blob/develop/rebar.config#L1) explicitly requires R16 or R17 and all 2.0 development was done against R16 so it is a good idea to be explicit at the top-level.
- [x] Update riak_ee repo with this change


seancribbs:
:+1: Easy-peasy

kellymclaughlin:
PR for riak_ee: https://github.com/basho/riak_ee/pull/250

PR #562, opened 2014-06-20
Change DEVNODES=5 to DEVNODES=8 in Makefile

andrewjstone:
Some SC tests require 8 nodes. Therefore raise the default to prevent
spurious errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ relclean:
 ##    make stagedevrel DEVNODES=68
 
 .PHONY : stagedevrel devrel
-DEVNODES ?= 5
+DEVNODES ?= 8
 
 # 'seq' is not available on all *BSD, so using an alternate in awk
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
:+1:

jaredmorrow:
Why was this changed here? We already have the notion of devnodes
overridden in our riak_test scripts.

andrewjstone:
@jaredmorrow AFAIK the riak_test scripts don't build the devrels required though, they just copy them. The error that occurs with less nodes is also cryptic. Can't write app.config essentially. This was actually a suggestion from @kellymclaughlin after I wasted some time on this today. I can revert it, but I'd like to prevent others from running into this issue another way if we do that.

jaredmorrow:
That's fine, I was just curious.  The scripts in riak_test do call 'make
devrel' if you are running them to bootstrap, but you are correct that the
current devrel stuff does not.

PR #571, opened 2014-07-22
Add release notes for security

macintux:
- Document security, especially important caveats
- Some unintentional whitespace cleaning per Emacs


jaredmorrow:
+1, please can you cp this over to riak_ee / 2.0 after this merges?  The files should be identical.

macintux:
Should be identical but aren't. Will do.

PR #572, opened 2014-07-23
Fix poorly-formatted lists

macintux:
Two lists were (I assume) unintentionally turned into literal blocks of text via indentation. Fix.


jaredmorrow:
+1, please merge to riak_ee at your leisure

PR #580, opened 2014-08-01
Add known issue for LWW Registers in Riak Maps

lenary:
basho/riak_dt#100 - mentioned in the mumble today

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -97,7 +97,7 @@ the context tells Riak what you've actually seen. All of the official
 Basho clients, with the exception of the Java client, handle opaque
 contexts for you. Please see the documentation for more details.
 
-Please see **Known Issues** below for two known issues with Riak maps.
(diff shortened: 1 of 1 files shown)


seancribbs:
:+1:

russelldb:
Disagree -1

russelldb:
If the LWW Reg was a key in Riak, does "time go backwards" if you write K={a, 2}, delete K, K={b, 0} ?

seancribbs:
@russelldb I think there's a subtle difference between "wrong" and "surprising", but from the user's perspective both could be "wrong". Let's discuss more in detail Monday as a group.

russelldb:
ok, thanks.

seancribbs:
Closing. To summarize:

The proper semantics don't really exist, we are figuring them out. Once we have determined them, the change can be made. Making this a Known Issue implies we know or at least have vague idea of what needs fixing. The reset-remove semantic can indeed cause things to "go back in time", but we know that LWW is lossy. _Caveat implementor!_

PR #614, opened 2014-10-24
riak-debug - check /etc/default/riak (ulimit should be set there)

binarytemple:
http://docs.basho.com/riak/1.4.10/ops/tuning/open-files-limit/#Linux

It is very common for Ubuntu users to forget to set ulimit -n 66xxx in /etc/default/riak 

This adds an additional check for this file to riak-debug

### DIFF
--- a/rel/files/riak-debug
+++ b/rel/files/riak-debug
@@ -373,6 +373,8 @@ if [ 1 -eq $get_syscmds ]; then
 
     # Dump files
     [ -f /etc/release ] && dump release cat /etc/release
+    [ -f /etc/default/riak ] && dump etc_default_riak cat /etc/default/riak
+    [ ! -f /etc/default/riak ] && dump etc_default_riak echo "error: /etc/default/riak missing"
     [ -f /etc/redhat-release ] && dump redhat_release cat /etc/redhat-release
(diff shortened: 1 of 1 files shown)


binarytemple:
@jpease does this sound like a good idea? Originated in ticket 9129 - wasted two CSE days.

pyrrho:
No reason not to do this.
Because /etc/default/riak is an Ubuntu-specific thing, I'd say changing the second line to

```
[ `uname -a | grep -i "ubuntu" -c` -gt 0 ] && [ ! -f /etc/default/riak ] && dump etc_default_riak echo "error: /etc/default/riak missing"
```

kesslerm:
Checking for distribution specific, undocumented strings is usually a bad idea. I'd just leave out the second line altogether. If the config file exists, collect it, otherwise not. Also, apart from Ubuntu, there are other distributions that make use of `/etc/default`, and the equivalent `/etc/sysconfig` exists on RedHat and SuSE.

pyrrho:
I was under the impression that only Ubuntu or Ubuntu-based systems made use of `/etc/default`. As that seems to be an incorrect assumption, I agree that the second line should be left.

PR #623, opened 2014-11-19
Remove afunix from EXOMETER_PACKAGES.

uwiger:
Afunix is not used at this time, and may cause build issues on some
platforms.


jburwell:
:+1: Performed a local `make distclean deps compile`, and the afunix dependency was not pulled and all modules compiled successfully.

PR #665, opened 2015-01-14
Pass flags through to the cluster status CLI command

nickelization:
This allows us to use the --help/--format flags on the cluster status command (and whatever other future flags may be implemented). Clique will filter out any invalid flags and display appropriate error messages anyway, so we don't need to be managing that in the riak-admin script.


andrewjstone:
:+1:

Issue #659, opened 2015-01-01
Riak 2.0.1-2.0.4 tags have deps pointing to riak_core '2.0' branch, not a tag


Vagabond:
Riak 2.0.0 was done correctly, but all subsequent tags have not pinned all dependancies to tags, so repeatable builds aren't currently happening. The addition of 'clique' as an incorrectly specified dependancy of riak_core on the riak core 2.0 branch is breaking all builds of riak from any tag after 2.0.0.

This is a very bad thing.

Vagabond:
So, for 2.0.0:

```
$ rebar list-deps | grep BRANCH
rebar_lock_deps_plugin BRANCH master git://github.com/seth/rebar_lock_deps_plugin.git
```

For 2.0.1:

```
$ rebar list-deps | grep BRANCH
cuttlefish BRANCH 2.0 git://github.com/basho/cuttlefish.git
cuttlefish BRANCH 2.0 git://github.com/basho/cuttlefish.git
cuttlefish BRANCH 2.0 git://github.com/basho/cuttlefish.git
eleveldb BRANCH 2.0 git://github.com/basho/eleveldb.git
cuttlefish BRANCH 2.0 git://github.com/basho/cuttlefish.git
riak_sysmon BRANCH 2.0 git://github.com/basho/riak_sysmon.git
riak_ensemble BRANCH 2.0 git://github.com/basho/riak_ensemble
eleveldb BRANCH 2.0 git://github.com/basho/eleveldb.git
riak_core BRANCH 2.0 git://github.com/basho/riak_core.git
riak_core BRANCH 2.0 git://github.com/basho/riak_core.git
sidejob BRANCH 2.0 git://github.com/basho/sidejob.git
bitcask BRANCH 1.7 git://github.com/basho/bitcask.git
riak_pipe BRANCH 2.0 git://github.com/basho/riak_pipe.git
riak_api BRANCH 2.0 git://github.com/basho/riak_api.git
riak_dt BRANCH 2.0 git://github.com/basho/riak_dt.git
riak_kv BRANCH 2.0 git://github.com/basho/riak_kv.git
merge_index BRANCH 2.0 git://github.com/basho/merge_index.git
riak_core BRANCH 2.0 git://github.com/basho/riak_core.git
riak_kv BRANCH 2.0 git://github.com/basho/riak_kv.git
canola BRANCH 2.0 git://github.com/basho/canola.git
node_package BRANCH 2.0 git://github.com/basho/node_package.git
[...]

PR #669, opened 2015-01-21
Updating rebar.config to be set purely to tags/SHAs

gcymbalski:
As requested following conversations with Andrew Stone, here's a PR to pin deps to the latest tags available in subrepos. Be aware that work may be needed in those subrepos to actually bring in commits that are expected, as previous builds have been against branches.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -14,16 +14,16 @@
                ]}.
 
 {deps, [
-        {node_package, ".*", {git, "git://github.com/basho/node_package.git", {branch, "develop"}}},
-        {lager_syslog, "2.0.3", {git, "git://github.com/basho/lager_syslog.git", {tag, "2.0.3"}}},
-        {cluster_info, ".*", {git, "git://github.com/basho/cluster_info.git", {branch, "develop"}}},
-        {riak_kv, ".*", {git, "git://github.com/basho/riak_kv.git", {branch, "develop"}}},
-        {riak_search, ".*", {git, "git://github.com/basho/riak_search.git", {branch, "develop"}}},
-        {riak_control, ".*", {git, "git://github.com/basho/riak_control.git", {branch, "develop"}}},
+        {node_package, ".*", {git, "git://github.com/basho/node_package.git", {tag, "2.0.0"}}},
+        {lager_syslog, ".*", {git, "git://github.com/basho/lager_syslog.git", {tag, "2.0.3"}}},
+        {cluster_info, ".*", {git, "git://github.com/basho/cluster_info.git", {tag, "2.0.0"}}},
+        {riak_kv, ".*", {git, "git://github.com/basho/riak_kv.git", {tag, "2.0.2"}}},
+        {riak_search, ".*", {git, "git://github.com/basho/riak_search.git", {tag, "2.0.0"}}},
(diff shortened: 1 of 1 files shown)


rebar.config
```diff
@@ -14,16 +14,16 @@
                ]}.
 
 {deps, [
-        {node_package, ".*", {git, "git://github.com/basho/node_package.git", {branch, "develop"}}},
-        {lager_syslog, "2.0.3", {git, "git://github.com/basho/lager_syslog.git", {tag, "2.0.3"}}},
-        {cluster_info, ".*", {git, "git://github.com/basho/cluster_info.git", {branch, "develop"}}},
-        {riak_kv, ".*", {git, "git://github.com/basho/riak_kv.git", {branch, "develop"}}},
-        {riak_search, ".*", {git, "git://github.com/basho/riak_search.git", {branch, "develop"}}},
-        {riak_control, ".*", {git, "git://github.com/basho/riak_control.git", {branch, "develop"}}},
+        {node_package, ".*", {git, "git://github.com/basho/node_package.git", {tag, "2.0.0"}}},
```
andrewjstone, line 17: `node_package` tag 2.0.0 doesn't actually contain the tip of the 2.0 branch that includes many new changes. We need to to create a new tag for the `node_package` repo on the 2.0 branch head. Then of course we'll have to ensure that changes get merged back to develop eventually. We also need to do the same thing for  any other repo that has commits later than the 2.0.0 tag. 

We then seriously need to consider our branching strategy on these repos. They don't have anywhere near the frequency of changes, or the formal release structure of riak/riak_ee/riak_core/riak_kv.

Usually jumping major and minor versions means significant change, but in our case it mainly means "matches riak/ee versioning." Since these versions change infrequently I suggest not developing on long lived versioned branches as much as we do now. For most changes we should probably just create a feature or bugfix branch directly off develop. Then we can simply merge that in and tag develop as part of the regular development process, updating rebar.config of parent projects as necessary. This will prevent the need to 'stabilize' a release and figure out "what is in it" for all repos. It also means we don't have to re-branch/re-tag each subrepo for each major riak release. 

Of course for repos with major changes and large, long running and diverging features we can create versioned branches as necessary.

andrewjstone:
After thinking about this PR for a bit, I think we should wait until the next release cycle to perform this versioning. This PR should probably just be closed along with it's ee brethren. I think there was probably a misunderstanding in what I was asking for. We need to version every repo properly and ensure we can have repeatable builds.  I'm in the process of writing a document to describe how we should do things for the next build cycle.

kuenishi:
@andrewjstone Maybe you just missed https://github.com/basho/riak/issues/659 discussion? Also, +1 to not using 2.0 but just develop, until we start working on 2.1 or even 3.0.

andrewjstone:
@kuenishi I did forget about that issue. The document I'm working on addresses the same concerns in a systematic manner. I don't think we should wait too long to address the failures in our release process. I just don't think we should try to get it in by Friday for 2.0.5. IMO we shouldn't release anything after 2.0.5 until the build process is documented, standardized internally and repeatable.

Basho-JIRA:
See comments here for additional detail - https://github.com/basho/riak/pull/669

These issues should just be closed without merging and we should just use the current release strategy (for now).

Closing the bug as we will not include the updates. 

**_[posted via JIRA by Derek Somogyi]**_

PR #676, opened 2015-02-04
riak-debug update

pyrrho:
This patch includes improvements, bug fixes, extended data collection.
It adds:
- A verbose failure flag to let us know when a `dump` command errors
  out (and what that error message is)
- `du` and `ls -lhR` metrics to the LevelDB and Bitcask backend
  metrics
- Anti-Entropy LOG and LOG.old collection
- The entire basho-patches directory
- The entire generated.config directory
- The kern.log file
- The current Java version
- The output of the /stats endpoint
- The output of `zfs list` and `zpool list`
- Information regarding the yokozuna directories, as well as all
  loaded schemas (controlled with a -y|--yokozuna option)
- A --ssl-certs flag to conditionally capture potentially private
  ssl certificates, as well as support for a RIAK_EXCLUDE environment
  variable that will allow riak-debug to skip sensitive files.

It fixed:
- `awk 'BEGIN {FS=""}` breaking on non-GNU systems (replaced with
  `cut -c1`)
- `find -maxdept` breaking on non-GNU systems (replaced with more
  generic solutions)
- Multi-backend relative paths failing to resolve on some systems

It should be noted that dpb-WIP-riak-debug-improvements separates these fixes/additions into individual commits. I might be easier to follow the changes there.

### DIFF
--- a/rel/files/riak-debug
+++ b/rel/files/riak-debug
@@ -78,7 +78,12 @@ dump () {
     if [ 0 -eq $retval ]; then
         printf '.' 1>&2
     else
-        printf 'E' 1>&2
+        if [ $verbose_output -gt 0 ]; then
+            printf 'Command '\'"$*"\'' failed. Continuing.' 1>&2
+            echo '' 1>&2 # Cheap, easy, guaranteed newline.
+        else
+            printf 'E' 1>&2
+        fi
     fi
 
     return $retval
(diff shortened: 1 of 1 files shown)


bsparrow435:
:+1:

bsparrow435:
Squash down to one commit and I'll merge.

PR #686, opened 2015-02-19
Pin meck dependency at top-level Riak.

cmeiklejohn:
This prevents situtations whre we have a race between libraries with
dependencies to see who pulls in meck, at what version, first.


andrewjstone:
:+1:  a8c3202

PR #693, opened 2015-02-24
Make sure that locked deps are obeyed for packaging

gcymbalski:
Make sure that Make calls rebar appropriately.

To test:
To test:
- Run make package! Observe output.
- Make sure that you don't see something like the below (as in previous builds)

> make[1]: Entering directory `/home/whatever'
> ./rebar get-deps

but instead

> ./rebar -C rebar.config.lock get-deps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,7 @@ get_dist_deps = mkdir distdir && \
                 git clone . distdir/$(CLONEDIR) && \
                 cd distdir/$(CLONEDIR) && \
                 git checkout $(REPO_TAG) && \
-                $(MAKE) deps && \
+                $(MAKE) locked-deps && \
                 echo "- Dependencies and their tags at build time of $(REPO) at $(REPO_TAG)" > $(MANIFEST_FILE) && \
                 for dep in deps/*; do \
                     cd $${dep} && \


cmeiklejohn:
:+1: looks good; built with the specified locked deps.

jaredmorrow:
This won't work for some packages because they callback into the makefile when they build.  You can fix it pretty easily, but will need to test it on all the platforms.

So checking in node_package to do due diligence, RPM/Deb/OSX are just fine, they call back into 'make rel' only (but you'll need to make sure that doesn't force a new `deps` requirement).

The following need to be changed to call only `make rel`:

https://github.com/basho/node_package/blob/develop/priv/templates/fbsd/Makefile#L156
https://github.com/basho/node_package/blob/develop/priv/templates/fbsdng/Makefile#L91
https://github.com/basho/node_package/blob/develop/priv/templates/solaris/Makefile#L29
https://github.com/basho/node_package/blob/develop/priv/templates/smartos/Makefile#L246

I'll leave testing my theory as an exercise to the reader ;)

gcymbalski:
Very cool, I was wondering if node_package would need tweaks. Thanks for the heads-up :)

jaredmorrow:
No problem, that would've been a really hard bug to track down.

PR #696, opened 2015-02-25
Bump rebar.

cmeiklejohn:
@andrewjstone required to correctly apply `platform_define` for some dependencies; part of the path to Erlang 17.


andrewjstone:
:+1:

PR #698, opened 2015-02-26
One more loose end on obeying locked deps

gcymbalski:
This was the last straggler, making all OSes obey locked deps properly. Woo!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ test: deps compile testclean
 ##
 ## Release targets
 ##
-rel: deps compile generate
+rel: locked-deps compile generate
 
 relclean:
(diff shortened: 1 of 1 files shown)


seancribbs:
So, my main reservation here is that this means a new lock file will be generated _every_ time a release is generated, even in development. Is there an alternative?

gcymbalski:
Oh, this just uses the locked deps instead of remaking it (lockdeps vs. locked-deps- clear, no?). So it shouldn't cause that, but I will definitely double-check that tomorrow.

cmeiklejohn:
:+1: this ensures the release is build with locked dependencies and does not alter or regenerate the file

PR #722, opened 2015-04-03
Riak and Riak_EE can have the same riak-debug

pyrrho:
Because of the way `dump` is written, there's no harm in including
`riak-repl` commands in Riak's `riak-debug`. With that in mind,
there's no reason to maintain two separate versions of `riak-debug`.


pyrrho:
This PR (and the related three) is comically out of date. Closing it to get it off my dashboard.

Issue #509, opened 2014-03-14
SSL options in vm.args make nodetool stop working [JIRA: RIAK-1579]


613038475:
Hello,
I found use ssl options in vm.args will let nodetool (riak start, riak-admin status, etc...) stop working.
The reason is env.sh only pick up name and cookie from vm.args then pass to nodetool.
Here is a patch to fix this problem.
https://gist.github.com/613038475/9546619

Also I found this option is not valid.
-proto_dist inet_ssl
it should be
-proto_dist inet_tls

bsparrow435:
I just confirmed this locally against develop. The fix is valid but we really should have a riak_test to validate this.

613038475:
That patch is not useable, it will make erl boot from "riak start" or "riak console" duplicate the ssl argument.

seancribbs:
Note that previously, this was slurped in from a separate file. Unfortunately you cannot pass init arguments from the command line into an escript; they have to be specified in the comment line after the shebang. This means that any disterl-related arguments have to be in a file that is referred to in the comment line, which is what was done in `nodetool` in the past. Fixing this issue would require changes to OTP.

tombriden:
This has recently been an issue for me, the attached patch doesn't appear to work with 2.0.5 but a slight variation on it works and there are no duplicate SSL variables when using "riak console" or "riak start"

```
SSL_ARGS=`grep -e '^\-ssl_dist_opt' -e '^\-proto_dist' $RUNNER_ETC_DIR/vm.args | tr '\n' ' '`
# Setup command to control the node
NODETOOL="eval ERL_FLAGS=\"$ERL_FLAGS $SSL_ARGS\" $ERTS_PATH/escript $ERTS_PATH/nodetool $NET_TICKTIME_ARG $NAME_ARG $COOKIE_ARG"
```

there's also a separate issue with some extra output from nodetool when using SSL making get_pid fail, fixable with

```
PID=`$NODETOOL getpid < /dev/null|head -1`
```

ryanbamford:
have modded this additionally to include a check for the file 

if [ -f $RUNNER_ETC_DIR/vm.args ]; then
   SSL_ARGS=`grep -e '^\-ssl_dist_opt' -e '^\-proto_dist' $RUNNER_ETC_DIR/vm.args | tr '\n' ' '`
[...]

PR #728, opened 2015-04-15
Reference NODETOOL by command alias (basho/riak#509)

tombriden:
depends on change to node_package: basho/node_package/pull/183


bsparrow435:
Going to need to fix this without changing nodetool to use aliases as this will break scripts used by other projects who also use nodetool.

PR #734, opened 2015-05-01
Make default handoff_ip value 0.0.0.0 in vars.config

JeetKunDoug:
so nodes don’t hand off to themselves in production clusters. Also update dev_vars.config.src to set handoff_ip to 127.0.0.1 for devrel builds.


seancribbs:
:+1: Identical to EE

PR #736, opened 2015-05-02
Bugfix/2.1.1 use locked deps for dist and fix deps to 2.1.0

jonmeredith:
Apply the build fix to use locked-deps when building the release tarball/packages that was applied to 2.0 after it diverged from master (needs to be added to develop too). Otherwise the rebar.config settings are used.

Reconstruct the locked deps as of the released 2.1.0 tarball so that the 2.1.1 release only differs in the top level packaging to pick up the handoff.ip changes for cuttelfish.

Companion PR to basho/riak_ee#334

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ test: deps compile testclean
 ##
 ## Release targets
 ##
-rel: deps compile generate
+rel: locked-deps compile generate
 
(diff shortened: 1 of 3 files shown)


andrewjstone:
+1 7bb76dd

PR #743, opened 2015-06-11
Fix issues with scripts when CDPATH is set

jvoegele:
The CDPATH environment variable can cause the cd command to sometimes
produce output. This breaks `runner_script_dir` detection in several
scripts. This fix suppresses the output from the cd command whenever it
is produced and allows the scripts to successfully detect the correct
directory for `runner_script_dir`.

For more information on when cd might produce output, see:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cd.html#tag_20_14_10

### DIFF
--- a/rel/vars.config
+++ b/rel/vars.config
@@ -41,7 +41,7 @@
 %%
 %% bin/riak
 %%
-{runner_script_dir,  "\`cd \\`dirname $0\\` && /bin/pwd\`"}.
+{runner_script_dir,  "\`cd \\`dirname $0\\` 1>&- && /bin/pwd\`"}.
 {runner_base_dir,    "{{runner_script_dir}}/.."}.
 {runner_etc_dir,     "$RUNNER_BASE_DIR/etc"}.
 {runner_log_dir,     "$RUNNER_BASE_DIR/log"}.
--- a/rel/vars.config
+++ b/rel/vars.config
@@ -41,7 +41,7 @@
 %%
 %% bin/riak
 %%
-{runner_script_dir,  "\`cd \\`dirname $0\\` 1>&- && /bin/pwd\`"}.
+{runner_script_dir,  "\`cd \\`dirname $0\\` 1>/dev/null && /bin/pwd\`"}.
 {runner_base_dir,    "{{runner_script_dir}}/.."}.
 {runner_etc_dir,     "$RUNNER_BASE_DIR/etc"}.
 {runner_log_dir,     "$RUNNER_BASE_DIR/log"}.
--- a/rel/vars/dev_vars.config.src
+++ b/rel/vars/dev_vars.config.src
@@ -42,7 +42,7 @@
 %%
 %% bin/riak
 %%
-{runner_script_dir,  "\`cd \\`dirname $0\\` 1>&- && /bin/pwd\`"}.
+{runner_script_dir,  "\`cd \\`dirname $0\\` 1>/dev/null && /bin/pwd\`"}.
 {runner_base_dir,    "{{runner_script_dir}}/.."}.
 {runner_etc_dir,     "$RUNNER_BASE_DIR/etc"}.
 {runner_log_dir,     "$RUNNER_BASE_DIR/log"}.
--- a/rel/vars/dev_vars.config.src
+++ b/rel/vars/dev_vars.config.src
@@ -42,7 +42,7 @@
 %%
 %% bin/riak
 %%
-{runner_script_dir,  "\`cd \\`dirname $0\\` && /bin/pwd\`"}.
+{runner_script_dir,  "\`cd \\`dirname $0\\` 1>&- && /bin/pwd\`"}.
 {runner_base_dir,    "{{runner_script_dir}}/.."}.
 {runner_etc_dir,     "$RUNNER_BASE_DIR/etc"}.
 {runner_log_dir,     "$RUNNER_BASE_DIR/log"}.
--- a/rel/vars/perf_vars.config.src
+++ b/rel/vars/perf_vars.config.src
@@ -49,7 +49,7 @@
 %%
 %% bin/riak
 %%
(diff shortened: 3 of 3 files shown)


lukebakken:
Would it be simpler to just add `unset CDPATH` at the top of the scripts?

https://github.com/sstephenson/rbenv/pull/326

or, `unset CDPATH` and this, if necessary:

https://github.com/sstephenson/rbenv/commit/e3f72ebae20768079ca4b4425a364900f3f16fc6

jvoegele:
That might indeed be simpler, but it cannot be isolated in the same way that the current approach can. In other words, to unset the `CDPATH` would require doing so in each script that uses the `runner_script_dir` variable, whereas if we fix the problem in the `runner_script_dir` itself it applies to any and all scripts that use it.

What do you think?

lukebakken:
I vote for fixing this "once and for all" in the top of all of our scripts. It may be more work now, but it would prevent this issue permanently going forward.

JeetKunDoug:
So two issues - We're not sure if `unset CDPATH` is sufficient, and the other proposal, `export -n CDPATH` isn't actually POSIX/portable from what I can tell. We do, however, know that this fix (redirecting cd's stdout) will _always_ work, no matter what stupid thing some shell decides to add to cd to output additional things.
-EDIT- Also, editing the scripts means this fix also affects node_package, not just riak

lukebakken:
[`unset CDPATH` is sufficient](http://pubs.opengroup.org/onlinepubs/009696699/utilities/unset.html), I shouldn't have linked to the other bit of code.

It should just be a matter of adding `unset CDPATH` here, unless I'm missing something:

https://github.com/basho/node_package/blob/develop/priv/base/env.sh#L17

JeetKunDoug:
also needs to be included in https://github.com/basho/node_package/blob/develop/priv/base/runner#L6 if we go that way.

JeetKunDoug:
More scripts which would need unsets:
- https://github.com/basho/riak/blob/89e4df90026966838e3156df51688c1b0dbcf3c5/rel/files/search-cmd#L6
- https://github.com/basho/riak/blob/89e4df90026966838e3156df51688c1b0dbcf3c5/rel/files/riak-admin#L5
- https://github.com/basho/riak/blob/89e4df90026966838e3156df51688c1b0dbcf3c5/rel/files/riak-debug#L137
- https://github.com/basho/riak_ee/blob/8ef6406f354f323ec627dae0859ff46da55d9251/rel/files/search-cmd#L5
- https://github.com/basho/riak_ee/blob/8ef6406f354f323ec627dae0859ff46da55d9251/rel/files/riak-repl#L6
- https://github.com/basho/riak_ee/blob/8ef6406f354f323ec627dae0859ff46da55d9251/rel/files/riak-admin#L6
- https://github.com/basho/riak_ee/blob/8ef6406f354f323ec627dae0859ff46da55d9251/rel/files/riak-debug#L137

lukebakken:
Well that's unfortunate. I'd like to suggest that `1>/dev/null` be used instead of `1>&-` for [this reason](http://stackoverflow.com/questions/14654577/difference-redirecting-to-bit-bucketaka-dev-null-vs-closing-the-file-descript) as well as the `1>/dev/null` idiom being more familiar.

lukebakken:
Example of the latter:

``` bash
lbakken@brahms ~
$ export CDPATH="$HOME/Projects"
lbakken@brahms ~
$ cd basho
/home/lbakken/Projects/basho
lbakken@brahms ~/Projects/basho
$ cd
lbakken@brahms ~
$ cd basho 1>&-; echo $?
-bash: cd: write error: Bad file descriptor
1
```

jvoegele:
@lukebakken I'm happy to change it to `1>/dev/null` as long as that is portable to all of our supported systems.

lukebakken:
If that's not portable we have a bigger problem than `CDPATH` :smile:

Seriously, though, the `/dev/null` device is part of the POSIX standard.

jvoegele:
Ha, fair enough.

lukebakken:
Also, using `1>&-` won't work:

``` bash
lbakken@brahms ~
$ export CDPATH="$HOME/Projects"
lbakken@brahms ~
$ cd basho 1>&- && /bin/pwd
-bash: cd: write error: Bad file descriptor
```

lukebakken:
Wow, such bizarre behavior:

``` bash
lbakken@BRAHMS-WIN8PRO ~
$ export CDPATH="$HOME/Projects/basho"
lbakken@BRAHMS-WIN8PRO ~
$ cd riak-nodejs-client 1>&- && /bin/pwd
-bash: cd: write error: Bad file descriptor
lbakken@BRAHMS-WIN8PRO ~/Projects/basho/riak-nodejs-client (issues/lrb/gh-69 *=)
$
```

The working directory changed even though `cd` failed to write to stdout. Not what I would expect. As expected, `/bin/pwd` did not run which would cause issues since we're trying to capture its output.

JeetKunDoug:
Yea - it seems as if cd doesn't actually care if it can write to stdout or not - it writes the failure to stderr but still succeeds. That said, > /dev/null is probably better (although I somehow still like 1>&- better :) )

lukebakken:
To be pedantic, we can't use `1>&-` - see my latest example.

JeetKunDoug:
Ah yea - I think it only works because we _happened_ to test it from the "right" parent dir to make it work even though the command returned no output. Not so good. /dev/null it is.

jvoegele:
I have made the change to redirect to `/dev/null` in a0453c3.

jvoegele:
Also: https://github.com/basho/riak_ee/pull/338

JeetKunDoug:
:+1:

JeetKunDoug:
Tested on Linux (Ubuntu) and Mac OS, for paranoia's sake. :+1:

jonmeredith:
The truly paranoid test with Solaris /bin/sh ;)

Issue #748, opened 2015-06-18
Riak-debug throws error during execution  [JIRA: RIAK-1874]


jenafermiller:
The `riak/rel/files/riak-debug` file throws the following error during execution: 

```
........E......EE......................dev2/bin/riak-debug: line 863: [: too many arguments
```

This error originates from the following line:

```
    if [ -n $exclude ]; then
```

To fix this, the `$exclude` must be in quotation marks as follows:

```
    if [ -n "$exclude" ]; then
```

jenafermiller:
PR for fix:  https://github.com/basho/riak/pull/749

cmeiklejohn:
:+1: 09ed9f3, please open backports for 2.0 and 2.1 if necessary.

PR #750, opened 2015-06-22
Bug fix for issue #748, riak-debug error thrown during execution back ported to 2.0

jenafermiller:
The `riak/rel/files/riak-debug` file throws the following error during execution: 

```
........E......EE......................dev2/bin/riak-debug: line 863: [: too many arguments
```

This error originates from the following line:

```
    if [ -n $exclude ]; then
```

To fix this, the `$exclude` must be in quotation marks as follows:

```
    if [ -n "$exclude" ]; then
```

_The fix is applied in this PR._


jonmeredith:
-1 I think this scoops up a lot of other changes and we just want 09ed9f3e8ec5c2ea447894ba55ef022898b46313 reapplied to 2.0

PR #753, opened 2015-06-23
backporting riak debug fix to 2.1

jenafermiller:
The `riak/rel/files/riak-debug` file throws the following error during execution: 

```
........E......EE......................dev2/bin/riak-debug: line 863: [: too many arguments
```

This error originates from the following line:

```
    if [ -n $exclude ]; then
```

To fix this, the `$exclude` must be in quotation marks as follows:

```
    if [ -n "$exclude" ]; then
```

_The fix is applied in this PR._


zeeshanlakhani:
:+1: ed316a4a655ba3556589190c963b745e69803866

PR #764, opened 2015-07-10
update rebar config for riak for 2.1 maint branches and tag updates

zeeshanlakhani:
Update 2.1's rebar.config to match our current versioning focus, remove need for meck at the top-level, matching riak_ee.


JeetKunDoug:
:+1:

PR #788, opened 2015-10-29
Moved riak_ql to reltool ref

cuyler:
This removes the riak_ql dependency from riak_kv, in favor of a modification to load riak_ql via reltool.config.


hazen:
:+1: 3e0a13a

Issue #796, opened 2015-12-02
Riak-debug fails on stock Solaris 10 [JIRA: RIAK-2314]


kesslerm:
On behalf of a customer:

riak-debug on solaris 10 isn't working...

```
riak-node:/opt/riak/etc root$ /opt/riak/bin/riak-debug 
..............................E....Eawk: syntax error near line 1 
awk: illegal statement near line 1 
sed: Extra text at end of command: s/^"//" -e "s/" /:/ 
Esed: Extra text at end of command: s/^"//" -e "s/".*$// 
/opt/riak/bin/riak-debug[507]: test: argument expected 
/riak_core_ring* not found 
E.Eawk: syntax error near line 1 
awk: illegal statement near line 1 
sed: Extra text at end of command: s/^"//" -e "s/".*$// 
/opt/riak/bin/riak-debug[540]: test: argument expected 
sed: Extra text at end of command: s/^"//" -e "s/".*$// 
/opt/riak/bin/riak-debug[546]: test: argument expected 
awk: syntax error near line 1 
awk: illegal statement near line 1 
sed: Extra text at end of command: s/^"//" -e "s/".*$// 
/opt/riak/bin/riak-debug[614]: test: argument expected 
.sed: Extra text at end of command: s/^"//" -e "s/".*$// 
sed: Extra text at end of command: s/^"//" -e "s/".*$// 
sed: Extra text at end of command: s/^"//" -e "s/".*$// 
/opt/riak/bin/riak-debug[794]: test: argument expected 
./tmp/riak@node-riak-debug: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
/tmp/riak@node-riak-debug/config: bad substitution 
.tar: z: unknown function modifier 
[...]

PR #797, opened 2015-12-04
Support riak-debug on Solaris 10

kesslerm:
A fix for #796 (RIAK-2308) (RIAK-2314).
- The default Solaris 10 version of awk doesn't support gsub, use xpg4 awk (nawk) instead.
- Tar on Solaris 10 has no support for creating compressed tar.gz files. Pipe tar file into gzip instead.
- Non-bash (e.g. ksh) shells may not have support for single instances of double quotes nested in single quotes. Escape nested double quotes.


kesslerm:
Has been merged by separate PRs to 2.0, 2.1 and develop 2.2 branches. Still not merged back to develop, but nobody cares(?).

PR #798, opened 2015-12-04
Support riak-debug on Solaris 10 (2.0 branch)

kesslerm:
A fix for #796 (RIAK-2308) (RIAK-2314).
- The default Solaris 10 version of awk doesn't support gsub, use xpg4 awk (nawk) instead.
- Tar on Solaris 10 has no support for creating compressed tar.gz files. Pipe tar file into gzip instead.
- Non-bash (e.g. ksh) shells may not have support for single instances of double quotes nested in single quotes. Escape nested double quotes.

### DIFF
--- a/rel/files/riak-debug
+++ b/rel/files/riak-debug
@@ -492,7 +492,7 @@ if [ 1 -eq $get_riakcmds ]; then
     riak_epaths=`make_app_epaths "${riak_app_config}"`
 
     # Get one http listener (epath might output a newline-separated list).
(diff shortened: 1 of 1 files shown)


JeetKunDoug:
:+1: tested on Centos (bash + ksh), Mac OS X (bash), and Solaris 10 (bash and /sbin/sh, the default shell, where this failed before the fixes)

kesslerm:
@JeetKunDoug was this ever merged forward into the 2.1 branches of Riak and Riak-EE?

JeetKunDoug:
Not yet, but will be before 2.2 (was too late for 2.1.2/3 release).

PR #809, opened 2016-02-22
don't leave bin/riak-shell behind when making rel

hmmr:
Add a new entry in reltool.config to fetch deps/riak_shell/bin/riak-shell escript for release and packaging.


hazen:
I've already got a directive to include `riak-shell`: https://github.com/basho/riak_ee/blob/riak_ts-rc-1.2.0/rel/reltool.config#L149

The script needs to be run through the preprocessor to do some constant replacements, so it must be a template.

hmmr:
@javajolt you are correct. Fixed and force-pushed.

hazen:
Thanks, however 1.2 is EE only which has this change.  We'll need to same one on `riak` for 1.3, however.

PR #810, opened 2016-02-23
don't leave behind riak-shell and its schema when making rel

hmmr:


hazen:
We need more changes than that, I'm afraid: https://github.com/basho/riak_ee/pull/370

hazen:
Also need this change: https://github.com/basho/riak_ee/pull/372

PR #811, opened 2016-02-24
Bump Riak 2.0 cuttlefish version to 2.0.2p1

nickelization:
This pulls in a slightly updated version of cuttlefish that includes
Doug's fix for increasing the default ERL_MAX_PORTS setting.


JeetKunDoug:
:+1: 4f11213

PR #822, opened 2016-03-28
Use a different method to determine the REPO name from the tag rather…

ooshlablu:
… than hard coding it into the makefile.


JeetKunDoug:
:+1:

PR #823, opened 2016-03-31
Update release notes for 2.1.4

lrnrthr:
The initial draft of the 2.1.4 release notes. I think there might be additional info we would want to include, but I wasn't sure whether the PAs covered it or not...


JeetKunDoug:
Superseded by release notes in #827

PR #837, opened 2016-05-17
Add pkg-rel to the Makefile

ooshlablu:


JeetKunDoug:
+1 99d59bd (not that there's a BORS here)

PR #858, opened 2016-07-12
Put version in the packaged devrels for riak test runs, update path f…

ooshlablu:
…or rels for MARQS


fadushin:
+1 e6ded3e

jbrisbin:
I don't remember for certain but the ones I created the first time I did this used a similar technique with the dot and it creates slight weirdness when unzipping. I switched to a splat instead and that worked better.

jbrisbin:
@ooshlablu +1 /cc @mdigan @sanmiguel

PR #866, opened 2016-08-02
Use node_package 3.0.1 instead of develop branch

nickelization:
In EE 2.0 we use the 3.0.1 tag, but for some reason in OSS 2.0 we're using the develop branch. This happens to be the same right now, but may change in the future, and we should use a tag like EE to keep things consistent. (For reference, the develop-2.2 branch is also using the 3.0.1 tag for both OSS and EE, so OSS 2.0 definitely seems like the odd one out.)

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -14,7 +14,7 @@
                ]}.
 
 {deps, [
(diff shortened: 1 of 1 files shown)


bsparrow435:
nice catch, this would have been really annoying to fix later. +1

PR #867, opened 2016-08-05
Added dist-query option to riak-admin

fadushin:


angrycub:
Opted instead to use the set command to set the cuttlefish setting via `riak-admin`.

PR #894, opened 2017-02-15
Remove the obsolete reformat-indexes command

nickelization:
This has not been needed since Riak 1.3; we're well past the point where we can delete the code and move on. Just to be sure though, I did run this by Client Services and they said the last time a ticket mentioned a customer using this command was in 2014.

This is the corresponding OSS PR to basho/riak_ee#418 (RIAK-3193). That one was already reviewed and merged, so I assume this should be a shoo-in.


fadushin:
+1

PR #896, opened 2017-02-22
Add release policy

ooshlablu:


How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
+This doc contains the internal Basho policy for building, testing, and releasing Riak.
+
+# Scheduling
+All public releases must conform to the following schedule:
+* An expected delivery date on a Wednesday. Each delivery date must have at least one fallback date that is no earlier than one week after the expected delivery date.
+* A Go/No-Go scheduled on the Thursday after the expected delivery date or fallback date to ensure that the public release can proceed as planned. During the Go/No-Go we must ensure:
+  * Giddyup has been triaged, and there are no major issues
+  * Smoke testing has completed successfully on all supported platforms
+  * Documentation content has been merged and triaged if necessary, and release notes are ready to ship
+  * Packages for all supported platforms have been built and are ready to stage
```
lrnrthr: Can we say "built, reviewed, and are ready to ship"? I think having Product/Docs review the packages has been really helpful in catching hiccups early.

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
+This doc contains the internal Basho policy for building, testing, and releasing Riak.
+
+# Scheduling
+All public releases must conform to the following schedule:
+* An expected delivery date on a Wednesday. Each delivery date must have at least one fallback date that is no earlier than one week after the expected delivery date.
```
lrnrthr: Can we clarify what "expected delivery" means? I read this a couple times before realizing it was not "expected delivery of the release to the public".

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+* A Go/No-Go scheduled on the Thursday after the expected delivery date or fallback date to ensure that the public release can proceed as planned. During the Go/No-Go we must ensure:
+  * Giddyup has been triaged, and there are no major issues
+  * Smoke testing has completed successfully on all supported platforms
+  * Documentation content has been merged and triaged if necessary, and release notes are ready to ship
+  * Packages for all supported platforms have been built and are ready to stage
+  * Any marketing announcements have been reviewed and are ready to ship
+* A final staging date on the Friday after the Go/No-Go. During this time, all packages will be pushed to their final download locations, the `External Release` docs will be sent around internall for review, and the indexes for the download pages will be generated
```
lrnrthr: What is an `External Release` doc?

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+* A public release date on the Tuesday after the final staging date. This date cannot overlap with another public release.
```
lrnrthr: Weren't we initially talking about trying to do two releases at once to see how it went? Has that become obviously a bad idea? If so, why? #curious

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+    product_version: 2.0.5
+    product: riak_ee
+```
+  * Tag riak_test with the same `annotated tag` that you will use for riak
+  * Push the tag to github  
+* Bundle Riak
+  * Lock the deps: `make lock`
+  * Tag riak with an `annotated tag`  
+* Giddyup runs will be kicked off automatically after the package builds complete successfully, but results must be triaged after each run. Giddyup runs usually complete 4.5 hours after the tag is initially received by the system.  
+* Inform the `Release Management` team that a new version of riak is being built, and schedule some time for Smoke testing.
```
lrnrthr: Are we defining who makes up the `Release Management` team?

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+  * Giddyup has been triaged, and there are no major issues
+  * Smoke testing has completed successfully on all supported platforms
+  * Documentation content has been merged and triaged if necessary, and release notes are ready to ship
+  * Packages for all supported platforms have been built and are ready to stage
+  * Any marketing announcements have been reviewed and are ready to ship
+* A final staging date on the Friday after the Go/No-Go. During this time, all packages will be pushed to their final download locations, the `External Release` docs will be sent around internall for review, and the indexes for the download pages will be generated
+* A public release date on the Tuesday after the final staging date. This date cannot overlap with another public release.
+
+# Pre-Release
+All releases of Riak will have at least one `Release Candidate` before the final package build. These RCs will contain all of the `code, backports, and bugfixes` required to build the Final Release, but will not have any updated `Release Notes`. Release Notes updates will be the only difference between RCs and the final version that will be released to the public.
```
lrnrthr: I don't know if it makes sense here, but I would love it if we noted somewhere official that Release Notes in packages = link to the URL of the release notes on the docs site.

lrnrthr (commented):
This looks really amazing, Paul!

I think most of my team's stuff comes after the stuff listed in Pre-Release, so let me know if/when you're ready and I'm happy to contribute :)

How-To-Release-Riak.md
hazen: Why Wednesday? Sounds like we are wanting to standardize the release days, etc, which I think is a great idea.

How-To-Release-Riak.md
hazen: I was also going to say, it would be nice to have the release URL already packaged with the RC, even if there is no content yet behind it

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+  * Package Build
+  * Internal Publishing
+  * riak_test (giddyup) runs
+  * Smoke Testing (on at least the final RC)
+
+The following procedure will cover all of these steps:  
+* Ensure `riak_test` is configured to automatically run after the package building steps:
+  * If there is no branch in `riak_test` that corresponds to `riak`, create it
+  * `basho_builds.yml` must exist in the top level of the repo, and contain the list of devrels that will be used to test various upgrade / downgrade scenarios during riak_test runs. The contents of the file will be similar to:
+ ```
```
hazen, line 47: Want to add a note about the basho_bench setting?

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+  * Any marketing announcements have been reviewed and are ready to ship
+* A final staging date on the Friday after the Go/No-Go. During this time, all packages will be pushed to their final download locations, the `External Release` docs will be sent around internall for review, and the indexes for the download pages will be generated
+* A public release date on the Tuesday after the final staging date. This date cannot overlap with another public release.
+
+# Pre-Release
+All releases of Riak will have at least one `Release Candidate` before the final package build. These RCs will contain all of the `code, backports, and bugfixes` required to build the Final Release, but will not have any updated `Release Notes`. Release Notes updates will be the only difference between RCs and the final version that will be released to the public.
+
+## Pre-release Procedure
+All Release Candidates will undergo the following release stages to ensure quality:
+  * Package Build
```
hazen: Want to mention that were are two different tags and repos: one for OSS and the other for EE?

How-To-Release-Riak.md
```diff
@@ -0,0 +1,56 @@
...
+  - version: 2.0.5
+    product_version: 2.0.5
+    product: riak_ee
+```
+  * Tag riak_test with the same `annotated tag` that you will use for riak
+  * Push the tag to github  
+* Bundle Riak
+  * Lock the deps: `make lock`
+  * Tag riak with an `annotated tag`  
+* Giddyup runs will be kicked off automatically after the package builds complete successfully, but results must be triaged after each run. Giddyup runs usually complete 4.5 hours after the tag is initially received by the system.
```
hazen: Is there a document on triaging Jenkins and or Ansible Tower? Not sure this is the right place for it, but a pointer might be nice since it's part of the release cycle.

How-To-Release-Riak.md
ooshlablu: @laurenrother yes, I'll fix it in the next iteration.

@javajolt Wednesday would be delivery from Eng, which would give us time to have the go/no-go the next day after we sit on it, and have a day to stage, etc. But yes, standardization, similar to what the TS team is doing

How-To-Release-Riak.md
ooshlablu: The review that I think is most helpful in that regard is when they are actually staged in the download locations, which I have as a different step. I want the go/no-go to happen before staging so I don't have to double up efforts if we back out at this point. I can explicitly call that out in the staging step.

How-To-Release-Riak.md
lrnrthr: Yeah, and that is 100% easy to do since it will always be the same URL just incremented up a number.

How-To-Release-Riak.md
ooshlablu: something like this: https://github.com/basho/internal_wiki/wiki/riak-2.0.8-External-Release

I don't know really what to call it, since it's for internal use only, but contains the links for the external packages. I'm open to suggestions.

How-To-Release-Riak.md
ooshlablu: We have to try it first and see, and if it works, we can change the policy. For now, I want to set expectations that we won't be doing it until we have done that experiment so it doesn't surprise us.... because you know it will :-)

How-To-Release-Riak.md
ooshlablu: @laurenrother I will change the wording so that becomes clear

@javajolt My concern is that the release notes are the only guaranteed way, at this point, to move the sha's forward so we can tag it and have jenkins pick it up. Everything except for 2.0 series relies on tagging the deps all the way down, which is probably the right way to do it, but since 2.0 just uses locked deps, we can't rely on it everywhere. Maybe we should just change that for 2.0 tho, and start tagging everything like we do for everything else. :man_shrugging:

How-To-Release-Riak.md
ooshlablu: Yeah, it's an actual team: @javajolt and @korry8911 
At some point soon, I'm gonna change this line anyway, since the smoke testing will be automated in jenkins as part of the standardized build process.

How-To-Release-Riak.md
ooshlablu: I was thinking that it would be evident in the doc I'm gonna drop in EE as well, but it probably makes more sense to call out each product in both places so there's no change of forgetting

How-To-Release-Riak.md
ooshlablu: Yeah, coming up shortly

ooshlablu: Should be a separate set of docs, but they don't exist yet. it would definitely be a good idea to link them in when they get put together.

How-To-Release-Riak.md
ooshlablu: yeah, coming up shortly

How-To-Release-Riak.md
angrycub: Let me start with, I love fallback dates.  Questions: Is a week _enough_ to smooth out all cases?  If we miss the first date, do we establish a new fallback once our fallback becomes the target?

ooshlablu: Yeah, I think we should establish a new one. I can add that in.

How-To-Release-Riak.md
hazen: We should just merge these two...

How-To-Release-Riak.md
hazen: It's a yuge team!

How-To-Release-Riak.md
```diff
@@ -0,0 +1,158 @@
...
+## Planning
+- [ ] Expected final-package delivery date has been scheduled
+- [ ] Fallback final-package delivery date has been scheduled
+- [ ] Go/No-Go has been scheduled
+- [ ] Fallback Go/No-Go has been scheduled
+- [ ] Public release date has been scheduled
+- [ ] Fallback Public release date has been scheduled
+- [ ] Product Management has issues a supported platforms list to DevOps and RelEng
+
+## Functional Complete
```
lrnrthr, line 122: Can we have a "predocs submitted to Tech Pubs team" added, please?

PR #902, opened 2017-03-13
extend schema for console and error log

joecaswell:
make console and error log size, rotation, and keep configurable in riak.conf

Currently these are hardcoded in the schema. Advanced.config requires
specifying a full handlers block that includes all settings for all log
files in order to modify them.

Add 6 configuration points:
log.console.size
log.console.rotation
log.console.rotation.keep
log.error.size
log.error.rotation
log.error.rotation.keep


rel/files/riak.schema
```diff
@@ -24,12 +24,54 @@
...
+
+%% @doc Maximum size of the error log in bytes, before it is rotated
+{mapping, "log.error.size", "lager.handlers", [
+  {default, "10MB"},
+  {datatype, bytesize}
+]}.
+
+%% @doc The number of rotated error log files to keep. When set to
+%% 'current', only the current open log file is kept.
+{mapping, "log.error.rotation.keep", "lager.crash_log_count", [
```
angrycub: should this be mapped to crash_log_count?

angrycub (changes requested):

joecaswell:
Absolutely correct, that should have been mapped to lager.handlers.  The way it was, if it were the only setting specified, it may be been ignored.  The tests passed because the translation functions pull the settings by the riak.conf name, not their mapped name.

angrycub (approved):

bsparrow435:
+1

PR #907, opened 2017-03-17
Update riak-debug shell script

pyrrho:
This is a pretty big changeset, but it looks larger than it actually
is. Numerous refactoring sweeps were made to homogenize comment and
coding styles, line lengths, code formatting, groupings and
orderings of code, etc., etc.. I'm guessing almost every line has
been touched, but significantly fewer have meaningfully changed.

Below is a list of changes I've made, translated from a personal
list I was keeping while making edit (made s.t. I could keep track
of what was done and what was left to do). There may be no
meaningful information for a review here, but I figured a reader
could choose to skip it if it where included. YMMV.

General changes:

- Updated the copyright date
- Added a descriptive comment block after the License flag
- Use a `use_defaults` flag for determining whether or not to use the
  default capture set, or if users provided a reduced list.
- Removed the `get_extras` section (it only included the vnode-status
  output. You want to see that, just run `riak-admin vnode-status`.)
- Added a verbose level one (printing failed commands) two (printing
  missing commands) and three (printing every command)
- Added a `poor_mans_newline` to make it easier for me to print
  newlines

Refactoring changes:

- 80 character line lengths (where possible)
- Reordered global variable definitions and command line parsing s.t.
  the two sections are more clearly separated
- Applied the `local` keyword to all function-local variables (really
  only applies do `dump`, but good practice is good)
- English punctuation on prosaic comments -- title blocks are still
  Title Case w/o significant punctuation.
- Modified and homogenized title formats (s.t. they're easier to see
  in my editor. Selfish coder is selfish.)
- Wrap shell variables ${in_squiggly_brackets}

Command changes:

- Added the below Riak commands to the `get_riakcmds` section
  - riak-admin cluster partitions
  - riak-admin cluster summary
  - riak-admin handoff details
  - riak-admin handoff summary
  - riak-admin ringready
  - riak-admin transfer-limit
  - riak-repl clustername
  - riak-repl nat-map show
  - riak-repl realtime cascades
- Removed the Cluster Info fetch from the `get_riakcmds` section (only
  applies to the OSS version of this file)
- Captured the `head` of any extant crash_dumps as part of the
  `get_logs` section
- Changed `lsb_release` to `lsb_release -a` (fixes basho/riak/#898)
- Added a `dmesg -T`, s.t. we can have timestamps on at least some
  platforms
- Added a series of calls to extract Bucket Type configurations

### DIFF
--- a/rel/files/riak-debug
+++ b/rel/files/riak-debug
@@ -1,10 +1,10 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 ## -------------------------------------------------------------------
 ##
 ## riak-debug: Gather info from a node for troubleshooting.
 ##
-## Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+## Copyright (c) 2017 Basho Technologies, Inc.  All Rights Reserved.
 ##
 ## This file is provided to you under the Apache License,
 ## Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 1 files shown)


nerophon:
All sounds good but why remove `riak-admin vnode-status`? Wasn't that missing for ages then we deliberately added it?

***update*** Drew notes that riak-admin vnode-status can badly bloat output; therefore it should be used only when specifically needed, and directly as per the notes above.

PR #927, opened 2017-11-07
Add repl dependency

martincox:
Added dependency for riak_repl and various rel config. Also reverted back to previous tag for YZ.


rebar.config
```diff
@@ -20,8 +20,9 @@
         {riak_kv, ".*", {git, "git://github.com/basho/riak_kv.git", {tag, "2.1.7"}}},
         {riak_search, ".*", {git, "git://github.com/basho/riak_search.git", {tag, "2.1.6"}}},
         {riak_control, ".*", {git, "git://github.com/basho/riak_control.git", {tag, "2.1.6"}}},
+        {riak_repl, ".*", {git, "git://github.com/basho/riak_repl.git", {tag, "2.1.8"}}},
```
cmeiklejohn, line 23: This repository isn't open source yet, thought?  Link is 404.

martincox: @cmeiklejohn It's about to be!

martincox: @cmeiklejohn open :)

russelldb:
Does this mean we're dropping SNMP and JMX?

russelldb:
Also, why not just take open the riak_ee repo and make that the _default_ Riak?

martincox:
No, this was just putting repl into open-source Riak. If there's an immediate need for it, let's add in the other components too. Any idea how widely used they are, if at all?

On the repo point, seemed less effort to add repl to to OSS Riak than to open up the private repo. It made more sense to keep it around - more issues, more activity, more forks, plain Riak over EE. There's not much in it really.

russelldb:
There's some missing vars from vars.config and dev_vars.config for the cluster manager port and IP. That aside it looks ok, but I haven't taken the time to build and test it yet.

Need to make a decision about jmx and snmp I guess. My feeling is that the next release should contain them, but deprecated, and then they can be removed after that.

russelldb:
Saying that, I added repl, jmx, and snmp to a OS Riak in the NHS fork and now Riak needs Java/Maven. Maybe you're right to add only repl.

hegy-100 (approved):

PR #930, opened 2017-11-16
Add riak_repl.schema.

firvida:
This way replication can be configured using riak.conf.  Partially reverts
8666c506954bd3b926a9b2c9d50b843c26b014fe.


russelldb:
Sorry, just saw this now. Do you know if this fixes https://github.com/basho/riak/issues/940 ?

If so, can you re-point the PR at develop-2.2?

ghost:
I wonder if we could go even further than this and make the advanced.config completely empty by default - the lager configuration section is, as far as I can tell, superfluous.  

It was [added](https://github.com/basho/riak/commit/16bc5ed2ecb52332e3d0c01398b4a5669e3fa233#diff-69d729a871f69bd71bda16acdd174f53) by @bsparrow435 back in 2016 but I can't see any reference to it anywhere in the code.

Everyone seems to be copy-n-pasting that block, but I'm smelling a potential cargo-cult.

ghost:
Nice fix +1

martincox:
The repl schema is copied over in develop-3.0 by way of the cuttlefish plugin schema discovery, so this isn't really required anymore.

PR #933, opened 2018-02-12
Extend mapdeps to add edge version labels

ghost:
Use (in case you forget) : 

```
make depgraph
```

```
Note: If you have nothing in deps/ this might be boring
Creating dependency graph...
Dependency graph created as riak.png
```

Before : 

riak

After : 

riak 1

### DIFF
--- a/misc/mapdeps.erl
+++ b/misc/mapdeps.erl
@@ -68,6 +68,15 @@ map_dir(BaseDir) ->
             exit(-1)
     end.
 
+
+% Obtain a nice version string to use in the label
+ver({tag,T}) ->
+   io_lib:format("tag:~s", [T]);
+ver({branch, B} ) ->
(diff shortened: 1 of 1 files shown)


misc/mapdeps.erl
```diff
@@ -92,7 +102,9 @@ map_rebar(BaseDir, Path, Acc) ->
                                            atom_to_list(DepName),
                                            "rebar.config"]),
                               map_rebar(BaseDir, DepPath, NA)
-                      end
+                      end;
+                 ({_DepName, _, {_,_,_}  }, A) ->
```
nickkeers, line 106: @binarytemple what's this function case for? Am i being dense? Looks like the above case will always match

ghost: Well spotted - let me remove that

PR #934, opened 2018-02-12
mapdeps.erl remove unused function clause

ghost:
Thanks @nickkeers for spotting !


ghost:
@martincox - left a dead function clause in the first attempt.

martincox:
I blame it on old eyes - that covers us both :laughing:

PR #943, opened 2018-04-24
Add pre-amble to release notes

russelldb:
I think it matters for posterity that this release is the first
post-Basho, particularily for those who are making a decision about
using Riak. To that end I added a little history to the release notes.

~~All that is left now if the known issues (@bryanhuntesl)~~


martinsumner:
+1

PR #945, opened 2018-04-24
Fixed spacing on h2 elements

smarj:
Markdown format fixes for some H2 elements.

I left the #Riak release notes bits alone, because it seems they shouldn't be H1s.


russelldb:
Appreciated, thanks @BillSmargiassi !

PR #950, opened 2018-07-31
Not all builders want to run all deps' tests

russelldb:
For example, some organisations do not use, nor have the time to
maintain, some deps, like riak_ensemble. This change adds an argument
`TEST_IGNORE` that will skip the eunit test running for the deps
assigned to the variable.

Example::

    make test TEST_IGNORE="riak_ensemble yokozuna"

Will run eunit, but skip those directories. Compile will still run.


Bob-The-Marauder:
+1

PR #959, opened 2019-03-08
fix mapdeps.erl

llelf:


ThomasArts (approved):
Alternatively, one can use `./rebar3 tree` which gives slightly more information, but the graphical output is superb ;-).

PR #964, opened 2019-03-29
Update leveled

martinsumner:
To handle GC issue.  This will make RC7.


martinsumner:
This change only impacts the memory management within leveled.  It has cleared smoke tests with leveled as a backend.  The standard NHS volume test is now running, but the first 3 hours have been consistent with past performance.

martinsumner:
volume test for RC7 was within 0.5% of previous tests on all measures. 

Full functional suite passed with leveled backend.

Release testing notes updated

PR #965, opened 2019-03-31
Add relx configuration for Riak.

cmeiklejohn:
Add a configuration for relx that uses the rebar3 cuttlefish plugin for properly building the configurations required schema files.

Cuttlefish compilation is still preventing the node from starting because cuttlefish doesn't run ```escript``` using the application libraries.  Therefore, any schema file that references a library in ```lib``` -- for instance, the Riak KV files reference a function in ```riak_core_throttle``` that is unavailable for cuttlefish unless the application libraries are loaded.

Another issue here is that in the configuration that uses a ```cd dirname $0 && /bin/pwd``` to derive the release path.  This is a problem because relx uses the escape functionality of mustache to conver this ```&&``` to ```&amp;amp;``` -- this is not a valid unix command.  Since there is no known way to prevent this escaping of a literal, this has been replaced by a ```;``` under the assumption that the directory change to ```dirname $0``` command will not fail.

Before this can be merged, the following issue has to be addressed:
https://github.com/vernemq/rebar3_cuttlefish/issues/35

cc: @martincox @martinsumner


llelf:
> Another issue here is that in the configuration that uses a cd dirname $0 && /bin/pwd to derive the release path. This is a problem because relx uses the escape functionality of mustache to conver this && to &amp;amp; -- this is not a valid unix command.

```sh
if cd "`dirname "$E"`"; then pwd; fi
```
?

cmeiklejohn:
It would need an else branch, and I'm not sure what's safe to use there?

llelf:
It doesn't need `else`, it is practically the same as `&&`.

ThomasArts:
This is outdated. We now have a release build in place.

PR #974, opened 2019-05-10
Add pre-start hook for checking node not running at start

ThomasArts:
Moved ulimit check from hooks/prestart to hooks, because one is free to run this script before status or any other command if one would like to do so. Is actually not bound to start.

It is debatable whether we really need 2 scripts for start check, but this is quite clean approach


martincox (approved):
I think it's nice to keep them separated. Seems cleaner to me. I don't foresee there being too many of these scripts, if it's a problem we can always consolidate later.

Good point on the prestart subdir - unnecessary hierarchy. :+1:

PR #975, opened 2019-05-10
Change the chkconfig script to 'nodetool eval' with inline erlang

martincox:
Seems cleaner than having a couple of files to handle a simple task - this is what we were trying to achieve
in Leeds.


martincox:
It bugged me that we had to have a couple of distinct scripts to achieve this, we can actually do it in a single script. What do you think @ThomasArts? We'd tried to do something similar on the first day, but couldn't quite figure it out.

ThomasAerts (approved):
Beautiful...  much better than having shell scripts hanging around

PR #976, opened 2019-05-13
Fixes for testing dependencies; travis

llelf:
Not all tests pass. At least some of them because they are being run as a different profile¹. Ideas?

¹) https://github.com/basho/riak/blob/8dcae40c548d2bed16a5d3dc60df78f72bebb02b/Makefile#L59

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: erlang
+otp_release:
+  - 20.3.8
+script:
+  - ./rebar3 as test compile
+  - make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,8 @@ language: erlang
 otp_release:
   - 20.3.8
 script:
-  - ./rebar3 as test compile
-  - make test
+  - ./rebar3 eunit
+
+  ## for deps tests:
+  # - ./rebar3 as test compile
+  # - make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: erlang
+sudo: true
 otp_release:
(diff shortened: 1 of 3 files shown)


ThomasArts:
The problem is not so much the profile as that some tests have hard coded paths. With the profile deptests the idea was to test each riak dependency separately. That works reasonable, as long as the paths are not wrong.

It seems really easy to add travis. Possibly we should add travis just for `./rebar3 eunit` to run the 5 unit tests (they pass). In the meanwhile, I work on getting the really failing tests and the tests that have wrong path to be fixed. After that, we add `make test` to this travis script. Agree?

llelf:
> It seems really easy to add travis

Yes, [not having to build a custom OTP in travis](https://github.com/basho/yokozuna/blob/develop-3.0/.travis.sh) helps tremendously ;-)

> The problem is not so much the profile as that some tests have hard coded paths

Custom profile doesn’t help either.  E. g. some dependencies have
`{erl_opts, [nowarn_export_all]}` for their test profile.


I changed it too only run riak’s eunit tests for now.

llelf:
Made more of dependencies tests work.

llelf:
These deps are failing (https://travis-ci.org/llelf/riak/builds/535391932)

a | b
-|-
cuttlefish | 
eper | https://github.com/basho/riak_kv/pull/1689 ?
hyper | 
ibrowse | 
kv_index_tictactree | 
lager | 
mochiweb | 
syslog | 
webmachine | 
yokozuna | https://github.com/basho/yokozuna/pull/757

PR #990, opened 2019-10-22
Switch non infinite rpc for cluster_info

martinsumner:
The rpc now has a 60s timeout which can be over-ridden with an environment variable.

Do not continue to use rpc_infinity (which isn't working)


martinsumner:
@ThomasArts

ThomasArts (approved):

PR #1013, opened 2020-03-20
Add ability to specify node base name in devrel and base port.

martincox:
Set the nodename prefix and base port for devrels.


martinsumner:
Is this still required?

Bob-The-Marauder (commented):
Although this is a nice toy, I too fail to see a major use for this.
Also, I don't see any port protection to avoid conflicts if the user specifies a port that clashes with other services. Admitedly, the original version didn't have that either but the hardcoded ports were chosen so that was rarely an issue.

PR #1014, opened 2020-03-25
Build TLS-enabled Riak node(s)

hanssv:
This is just proof of concept, but together with [riak_test/use_inet_tls](https://github.com/basho/riak_test/pull/1342) branch tests are being run against TLS-enabled Riak node(s). It remains to make this somewhat more user friendly/configurable.

Using TLS exposed a couple of problems with variables, and extension scripts, these were pre-existing but didn't manifest in an "all-default" setting.


hanssv:
DO NOT MERGE in current shape.

rel/files/riak-repl
```diff
@@ -6,6 +6,7 @@
 ORIGINAL_DIR=$(pwd)
 
 # Make sure CWD is set to runner run dir
+RUNNER_BASE_DIR={{runner_base_dir}}
```
martincox, line 9: Hey @hanssv, did something break that required this to be added? I think the runner_base stuff is overspill from the former node_package stuff, but isn't necessarily required any more, to my knowledge (needs a bit of cleanup to remove it).

hanssv: Yes, without this it runs in my home-directory, which means it does not find the SSL-config and thus can't talk to the node (or well it will refuse to even try to talk to it).

It wasn't necessary before (and it was horribly broken since bbmustache had HTML-encoded the bash command!)

PR #1018, opened 2020-04-29
Develop 3.0 292

martinsumner:
This tides up the makefile, and now `make test` will run unit tests across all repos.  This is now passing on OTP 22.

An extra step has been added to `make rel` to copy the release out of _build and into rel so that scripts based on building from source will continue to work as before.

All the dep updates now mean no gen_fsm_compat in this release.  There are a couple of outstanding warnings:

- bear, setup and exometer_core all have compilation warnings (brought in as part of the stats package - all repos managed by Ulf);
- basho_bench has a compilation warning (only required as a dep of yokozuna, and then only to setup riak_test);
- lager is excluded from the `make test` loop as test fail on OTP22, and this is not resolved in `erlang-lager`.

Currently this includes yokozuna - although a decision on whether to actually have yokozuna in the release is still to be made.


martincox (approved):
lgtm. 👍 

Out of curiosity, which scripts require the release to be copied?

martinsumner:
The `make rel` script in 2.9 creates the release in {RIAK_REPO}/rel/riak/ (e.g. etc, bin, log folders etc), whereas in OTP 20 + the rel is created in {RIAK_REPO}/_build/rel/rel/riak.  I just added a `cp -a` to the Makefile for `make rel` to copy this structure across to the old location.

martinsumner:
wait on https://github.com/basho/riak_kv/pull/1757

PR #1023, opened 2020-07-01
Mas 3.0 debianmake

martinsumner:
Significant changes to debian make process.  As rebar3_cuttlefish scripts are run as a pre-start hooks, it does not have root access to make links into running directory.  To resolve this the riak script has been split into two parts riakpre and riakrun:

riakpre is run as root
riakrun is run as riak

The cf_config script makes the links into a riak-controlled space (platform_gen_dir) and riakpre than makes the links from that place to the privileged directory configuration files.

When `sudo riak start` or `sudo service riak start` is called cf_config will create the initial links, but the actual config is read directly (not depending on the links).  the linked version of the config is only required when running `riak ping`, `riak stop` etc and these will use the linked version - and as they only make sense after a `riak start` it can be assumed the links have already been made by cf_config.

To simplify, and workaround issues with config merging across profiles, more configuration (including standard rel vars.config) has been moved into debian vars.config.  This leads to some repetition - but is easier to follow.  so now the debian release only uses the deb profile, and not the rel/deb profiles combined.


martinsumner:
This PR depends on this https://github.com/basho/rebar3_cuttlefish/pull/2

martincox (approved):

PR #1024, opened 2020-07-07
Bring RPM packaging inline with DEB

martincox:


martincox:
Also, changed the Makefile to remove locked-deps target, which was incorrectly calling `rebar3 upgrade`, meaning that when creating a release, the deps were being upgraded instead of using the lock.

Cleaned up the old packaging targets.

martinsumner (approved):
Confirmed that debian packaging remains OK, and tests still run as expected

PR #1025, opened 2020-07-08
script name is passed through from riak-admin to clique

martincox:
It is currently set to the base script, which is now `riak`. We changed some calls so that they
worked by suffixing the call with `-admin` to work around it.

There was an issue with running `riak admin handoff *`, which slipped through the
net. Looking at this again, seemed better to change the $SCRIPT var rather than
suffix the call, just in case there's anything else that we haven't caught, it
should account for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ $(eval devrel : $(foreach n,$(SEQ),dev$(n)))
 
 dev% : all
 	rel/gen_dev dev$* rel/vars/dev_vars.config.src rel/vars/$*_vars.config
-	$(REBAR) release -o dev/dev$* --overlay_vars rel/vars/$*_vars.config
+	$(REBAR) as dev release -o dev/dev$* --overlay_vars rel/vars/$*_vars.config
 
 stagedev% : all
 	rel/gen_dev dev$* rel/vars/dev_vars.config.src rel/vars/$*_vars.config
--- a/rebar.config
+++ b/rebar.config
@@ -128,8 +128,7 @@
                         [{custom, "hooks/riak_not_running"},
                         {custom, "hooks/check_ulimit"}]},
                     {post_start,
-                        [{pid, "/run/riak/riak.pid"},
-                        {wait_for_process, riak_core_node_watcher}]}
+                        [{wait_for_process, riak_core_node_watcher}]}
                     ]}
         ]}
     ]},
--- a/rebar.config
+++ b/rebar.config
@@ -121,14 +121,21 @@
                 ]}
         ]}
     ]},
-    {dev, [{relx, [{dev_mode, true}]}
+    {dev, [{relx, [
+            {dev_mode, true},
+            {extended_start_script_hooks,
+                    [{pre_start,
+                        [{custom, "hooks/riak_not_running"},
+                        {custom, "hooks/check_ulimit"}]},
+                    {post_start,
+                        [{pid, "/run/riak/riak.pid"},
+                        {wait_for_process, riak_core_node_watcher}]}
+                    ]}
+        ]}
     ]},
     {rpm, [
         {relx, [
             {overlay_vars, "rel/pkg/rpm/vars.config"},
-            {overlay, [
-                
-            ]},
             {overlay, [
                 {template, "rel/files/riak", "usr/bin/riak"},
                 {template, "rel/files/riakpre", "usr/bin/riakpre"},
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -9,7 +9,7 @@ ORIGINAL_DIR=$(pwd)
 cd $RUNNER_BASE_DIR
 
 # Identify the script name
-SCRIPT=`basename $0`
+SCRIPT="riak-admin"
 
 usage() {
     echo "Usage: $SCRIPT { cluster | join | leave | backup | restore | test | "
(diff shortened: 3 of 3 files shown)


martinsumner:
The riak-test `basic_command_line` fails on this branch:

```
09:24:08.162 [info] <0.170.0> Node is already up, `riak console` should fail
09:24:08.162 [info] <0.170.0> Running: /Users/martinsumner/rt/riak/current/dev/dev1/riak/bin/riak console
09:24:09.447 [info] <0.170.0> Exec: /Users/martinsumner/rt/riak/current/dev/dev1/riak/erts-9.3/bin/erlexec -boot /Users/martinsumner/rt/riak/current/dev/dev1/riak/releases/3.0/riak -mode embedded -boot_var ERTS_LIB_DIR /Users/martinsumner/rt/riak/current/dev/dev1/riak/lib -config /Users/martinsumner/rt/riak/current/dev/dev1/riak/sys.config -args_file /Users/martinsumner/rt/riak/current/dev/dev1/riak/vm.args -- console
Root: /Users/martinsumner/rt/riak/current/dev/dev1/riak
/Users/martinsumner/rt/riak/current/dev/dev1/riak
Protocol 'inet_tcp': the name dev1@127.0.0.1 seems to be in use by another Erlang node

09:24:09.447 [warning] <0.5.0> basic_command_line failed: {{assert,[{module,basic_command_line},{line,56},{expression,"rt : str ( ConsoleFail , \"Node is already running!\" )"},{expected,true},{value,false}]},[{basic_command_line,'-console_up_test/1-fun-2-',1,[{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},{line,56}]},{basic_command_line,console_up_test,1,[{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},{line,56}]},{basic_command_line,confirm,0,[{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},{line,38}]},{riak_test_runner,return_to_exit,3,[{file,"/Users/martinsumner/riak_test/src/riak_test_runner.erl"},{line,159}]}]}
09:24:09.447 [error] emulator Error in process <0.170.0> on node 'riak_test@127.0.0.1' with exit value:
{{assert,[{module,basic_command_line},{line,56},{expression,"rt : str ( ConsoleFail , \"Node is already running!\" )"},{expected,true},{value,false}]},[{basic_command_line,'-console_up_test/1-fun-2-',1,[{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},{line,56}]},{basic_command_line,console_up_test,1,[{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},{line,56}]},{basic_command_line,confirm,0,[{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},{line,38}]},{riak_test_runner,return_to_exit,3,[{file,"/Users/martinsumner/riak_test/src/riak_test_runner.erl"},{line,159}]}]}
09:24:09.447 [error] <0.5.0> 
================ basic_command_line failure stack trace =====================
{{assert,[{module,basic_command_line},
          {line,56},
          {expression,"rt : str ( ConsoleFail , \"Node is already running!\" )"},
          {expected,true},
          {value,false}]},
 [{basic_command_line,'-console_up_test/1-fun-2-',1,
                      [{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},
                       {line,56}]},
  {basic_command_line,console_up_test,1,
                      [{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},
                       {line,56}]},
  {basic_command_line,confirm,0,
                      [{file,"/Users/martinsumner/riak_test/tests/basic_command_line.erl"},
                       {line,38}]},
  {riak_test_runner,return_to_exit,3,
                    [{file,"/Users/martinsumner/riak_test/src/riak_test_runner.erl"},
                     {line,159}]}]}
=============================================================================
```

Was passing previously, but will double check that it is still passing on `develop-3.0`.

martinsumner:
Hmmm, this test is broken on develop-3.0 as well.

riak_console is failing to start as expected, but with an unexpected response.  The pre-start script should trigger [this response](https://github.com/basho/riak/blob/develop-3.0/rel/files/riak_not_running).

I note when testing riak console manually,  it is no longer possible to quit as q is not an option in the list of user switch commands:

```
User switch command
 --> h
  c [nn]            - connect to job
  i [nn]            - interrupt job
  k [nn]            - kill job
  j                 - list all jobs
  s [shell]         - start local shell
  r [node [shell]]  - start remote shell
  ? | h             - this message
```

martinsumner:
Ah.  I think I might have broken this by removing the pre-start hooks from the dev profile.

martincox:
Ahhh, is this caused by the change to prevent the VM from being stopped on a `riak attach`

martincox:
So it's just skipping the node up check that normally returns `node already started` or something like that?

martinsumner:
Yes.  the riak script produced by `make devrel` doesn't have the `riak_not_running` check as a pre-start hook.

I've added the extended_start_script_hooks to the `dev` profile, which I hoped would fix it, but it doesn't

martincox:
Those changes pass for me?!

martinsumner:
Hmm.  But not for me.  I noticed that the Makefile doesn't set a profile for make devrel, and so added an 'as dev':

```
$(eval devrel : $(foreach n,$(SEQ),dev$(n)))

dev% : all
	rel/gen_dev dev$* rel/vars/dev_vars.config.src rel/vars/$*_vars.config
	$(REBAR) **as dev** release -o dev/dev$* --overlay_vars rel/vars/$*_vars.config
```
But still doesn't work for me.  The generated riak script only has cf_config as a pre_start_hook.  Not sure what I'm getting wrong.

martinsumner:
Could you commit your changes to this branch, and I'll see if it works if I start fresh with this branch?

martincox:
Ahhh, so it's an oversight on the devrel, but not stagedevrel - I always use the latter, and guess you were using just devrel?

I'll push the changes shortly.

martincox:
@martinsumner pushed the change, if you want to give it another go

martinsumner:
So I understand now why I couldn't get the profile change to work ... I was looking at the wrong file.

I don't think `{pid, "/run/riak/riak.pid"}` should be in the post_start hooks for `dev`.

martincox:
> I don't think `{pid, "/run/riak/riak.pid"}` should be in the post_start hooks for `dev`.

Yep, saw I'd done that and removed 👍

martinsumner (approved):

PR #1026, opened 2020-07-08
Mas 3.0 riakrunpath

martinsumner:
There were three problems with the previous debian build:

- It failed on debian/ubuntu boxes where `/usr/sbin` was not in the path (by default Google cloud debian instances this is the case);

- There was an untidiness with requiring three scripts not one (e.g. riakpre, riak, riakrun);

- As pointed out by Steve Wood relx expects you to override RELX_CONFIG_PATH and VMARGS_PATH, not mess around with links to repoint the default path.

This update resolves these problems and moves back to a single riak script again.  this works on Ubuntu 20 and also debian 9/10 when when /usr/sbin is not in the path.


martinsumner:
@martincox - sorry, just when you thought we were all done ... I've gone and refined the build scripts again to go back to a single riak script (removing the riakpre/riakrun), and removing the need for the link to the link to the file.

This seems to solve the problem we were having with some debian/ubuntu setups where /usr/sbin was not in the path.

martincox:
@martinsumner no worries, I'll test the changes out in centos this morning 👍

martincox:
Looks good to me. Pushed a commit to remove the scripts from the specfile. All good 👍

martincox (approved):

PR #1035, opened 2020-08-21
Edit Makefile so package names include OTP version

Bob-The-Marauder:
Added OTP_VER to initial variables so that it can be included in the PKG_ID variable and thus the binary file name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ PKG_REVISION    ?= $(shell git describe --tags 2>/dev/null)
 PKG_BUILD        = 1
 BASE_DIR         = $(shell pwd)
 ERLANG_BIN       = $(shell dirname $(shell which erl 2>/dev/null) 2>/dev/null)
+OTP_VER          = $(shell echo $(ERLANG_BIN) | rev | cut -d "/" -f 2 | rev)
 REBAR           ?= $(BASE_DIR)/rebar3
 OVERLAY_VARS    ?=
 TEST_IGNORE     ?= lager riak basho_bench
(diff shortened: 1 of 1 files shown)


martinsumner:
Debian build doesn't like this:

```
debuild: warning:     debian/changelog(l1): version '3.0.1_OTP_22.3' is invalid: version number contains illegal ch
aracter '_'
```

Seems to work without the underscores, but then the name looks confusing because of the zero/oh confusion.

martinsumner:
Confirmed that it works fine on ubuntu

martincox:
@Bob-The-Marauder have you generated an RPM package with these changes? Something is sticking in my mind about rpmbuild being a bit iffy over hyphens in package names.

Bob-The-Marauder:
Please move to pending - I think there is a better way to do this but I haven't got it working yet.

In answer to the question, rpm build switches the hyphens to dots which does work but I don't like the outputted name.

In a perfect fix, `riak versions` should output `Riak KV 3.0.1 OTP 22.3` or equivalent thereof. My above solution does not do this yet.

Bob-The-Marauder:
**Small update**
I have been having a think about this in light of building the Yokozuna packages. The current Yokozuna package build process involves removing the comments in `rebar.config` to enable a Yokozuna build and then retagging before building so that it doesn't simply rebuild the original tag without the edits and drop Yokozuna as a result.

As future package builds will involve 4 varients for each OS (OTP 20.3, OTP 22.3, OTP 20.3 Yokozuna and OTP 22.3 Yokozuna), I'm going to have to update this Pull Request to include Yokozuna as well for distinguishing purposes. 

Short term, I may just look at updating my build scripts to include multiple local-retagging of the repo before building. This method allows me to include the OTP version and presense or absense of Yokozuna in the tag and hence the package name essentially for free.

Disclaimer: the edits made before tagging to make the packages are only to allow Yokozuna to be built. It is against my principles to include non-vetted code, even if I wrote it.

Bob-The-Marauder:
I discovered that the output of `riak versions` has nothing whatsoever to do with the changes here. As such, not proceeding with that part.

As to the Yokozuna code mentioned in my previous comment, I haven't finished it yet.

Given the above, please proceed with this PR as is. I'll submit a separate one once I have the Yokozuna code working.

martinsumner (approved):

Issue #1037, opened 2020-09-04
riak debug broken in Riak KV 3.0.1


Bob-The-Marauder:
**Expected behaviour KV 3.0.1:**
Execute `riak debug` and a debug.tar.gz file is created containing lots of information about the node.

**Actual behaviour KV 3.0.1**
```
root@buster:~# riak debug
Required file app_epath.sh not found. Expected here:
/usr/lib/riak/lib/app_epath.sh
See 'riak-debug -h' and manpage for help.
```

This has been tested with CentOS8, Debian Buster and Ubuntu Focal and Bionic with the same result for all of them.

martinsumner:
app_epath.sh used to be provided as part of [node_package](https://github.com/basho/node_package/blob/develop-3.0/priv/base/app_epath.sh), but we moved away from node_package to use relx.

If you want to get this working (or at least past this stage), you will need to add the script back into rel/files and add into rebar.config a relx overlay template to copy the script across to the desired location like https://github.com/basho/riak/blob/develop-3.0/rebar.config#L83-L95.  Use copy if there's no substitution required of standard variables, template if there is.

We came across riak debug not working during testing of 3.0, but as we don't use it, we didn't spend time fixing it.  Not sure problems there might be beyond the existence of this script, so how much work there would be involved in fixing it - but if it is just about correcting the existence of the script it should be straight forward.

Bob-The-Marauder:
Fixed in Riak KV 3.0.2 and higher

PR #1040, opened 2020-10-16
Riak-Debug fix

Bob-The-Marauder:
This fix addresses https://github.com/basho/riak/issues/1037 by doing the following:

- re-adding `app_epath.sh` from Riak KV 2.9.7 to `/usr/lib64/riak/lib` on rpm builds and `/usr/lib/riak/lib` on deb builds from `/rel/files/` via an edit to `rebar.config`.
- applying numerous fixes to `/rel/files/riak-debug` to render it compatible with KV 3.0.1.
- as `/usr/sbin/riak debug` actually executes `/usr/lib64/riak/bin/riak-debug -s` for reasons I could neither determine nor fix, I have made a work-around for this by adding a symlink from `riak-debug` to `/usr/sbin` in the post-install script for the packages.
- added a capture of directory structure for tictacaae and leveled.

The symlinks are only a temporary work-around. Once I have worked out why `riak debug` executes `riak-debug -s` and fixed it, I'll submit another pull request to update for that and remove the symlinks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@ PKG_REVISION    ?= $(shell git describe --tags 2>/dev/null)
 PKG_BUILD        = 1
 BASE_DIR         = $(shell pwd)
 ERLANG_BIN       = $(shell dirname $(shell which erl 2>/dev/null) 2>/dev/null)
-OTP_VER          = $(shell echo $(ERLANG_BIN) | rev | cut -d "/" -f 2 | rev)
 REBAR           ?= $(BASE_DIR)/rebar3
 OVERLAY_VARS    ?=
 TEST_IGNORE     ?= lager riak basho_bench
@@ -211,7 +210,7 @@ REVISION = $(shell echo $(REPO_TAG) | sed -e 's/^$(REPO)-//')
 
 # Primary version identifier, strip off commmit information
 # Changes to 1.0.3 or 1.1.0pre1 from example above
-MAJOR_VERSION	?= $(shell echo $(REVISION) | sed -e 's/\([0-9.]*\)-.*/\1/'|+="-OTP$(OTP_VER)")
+MAJOR_VERSION	?= $(shell echo $(REVISION) | sed -e 's/\([0-9.]*\)-.*/\1/')
 
 # Name resulting directory & tar file based on current status of the git tag
 # If it is a tagged release (PKG_VERSION == MAJOR_VERSION), use the toplevel
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ PKG_REVISION    ?= $(shell git describe --tags 2>/dev/null)
 PKG_BUILD        = 1
 BASE_DIR         = $(shell pwd)
 ERLANG_BIN       = $(shell dirname $(shell which erl 2>/dev/null) 2>/dev/null)
+OTP_VER          = $(shell echo $(ERLANG_BIN) | rev | cut -d "/" -f 2 | rev)
 REBAR           ?= $(BASE_DIR)/rebar3
 OVERLAY_VARS    ?=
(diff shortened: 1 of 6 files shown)


martinsumner:
I have no objections to this - no reason why it should impact anything else.

martincox:
@Bob-The-Marauder does riak-debug need to be symlinked to /usr/sbin? Or can it work in the same way as admin and repl via script extensions? I can't remember whether the script extensions do a node check before execution.

martincox:
Sorry, I'm asking questions that have already been answered (I should read the comments first). I'm curious as to why it doesn't work in the same way as admin and repl still, though. I can't say as I've ever used debug tbh - do we run it against a running node (in which case the node up check should be fine)?

martinsumner:
riaknostic is missing as a dependency (it got stripped out for 3.0).  so this needs to be added back in for `riak admin diag` to work.  Otherwise:

```
RPC to 'riak@127.0.0.1' failed: {'EXIT',
                                 {undef,
                                  [{riaknostic,main,[["--etc","--base"]],[]},
                                   {rpc,'-handle_call_call/6-fun-0-',5,
                                    [{file,"rpc.erl"},{line,197}]}]}}
```

martinsumner:
The riaknostic repo has dialyzer/xref failures on the `develop-3.0` branch so this needs some work before it can be added in.

martinsumner:
https://github.com/basho/riaknostic/commit/723c73dc877e1d1bb4e69f893255eb7c4ba2fdf3

martinsumner:
So there appear to be problems with the symbolic link:

```
martin_sumner@riak-ubuntu18:~/dbroot/riak$ sudo dpkg -i rel/pkg/packages/riak_3.0.2diag1-OTP22.3_amd64.deb 
Selecting previously unselected package riak.
(Reading database ... 202026 files and directories currently installed.)
Preparing to unpack .../riak_3.0.2diag1-OTP22.3_amd64.deb ...
Unpacking riak (3.0.2diag1-OTP22.3) ...
Setting up riak (3.0.2diag1-OTP22.3) ...
Installing new version of config file /etc/riak/riak.conf ...
ln: failed to create symbolic link '/usr/sbin/riak-debug': File exists
dpkg: error processing package riak (--install):
 installed riak package post-installation script subprocess returned error exit status 1
Processing triggers for ureadahead (0.100.0-21) ...
Processing triggers for systemd (237-3ubuntu10.42) ...
Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
Errors were encountered while processing:
 riak
```

martinsumner:
I suspect `dpkg -r` isn't removing the link, and so when I reinstall the install package fails on hitting the link

martinsumner:
There's more work to be done here.

I've managed to get riaknostic loading, without crashing riak.  However, not it can't load the vm.args:

```
RPC to 'riak@127.0.0.1' failed: {'EXIT',
                                 {function_clause,
                                  [{string,substr,
                                    ["/var/lib/riak/sys.config",1,-1],
                                    [{file,"string.erl"},{line,1909}]},
                                   {riaknostic_config,load_vm_args,0,
...
```

Presumably the `AppIndex - 1` is < 1.

This is caused by the fact that this is now a sys.config not an app.config file.  I will hack away for a bit longer to see if I can make this work.

martinsumner:
@Bob-The-Marauder Are you OK if we use this PR instead:

https://github.com/basho/riak/pull/1047

It contains modifications to this PR to get it to work on my machine.  I didn't find the symlink was necessary, and made some changes so that riaknostic would load and the 'riak admin diag' would run as expected.

I can now `dpkg -i ...` `sudo service riak start` then `sudo riak debug`.  Works on Ubuntu 18.

martinsumner:
Replaced by #1047

PR #1044, opened 2020-12-09
Fix/bt props cli

martincox:


martincox:
The escaped args sed is pulled from the old node_package runner script - it's necessary because of the quotes in the props string being stripped as the args are passed. One day we should replace json props entirely.

martinsumner:
There is a set of console tests in riak_test - https://github.com/basho/riak_test/blob/develop-3.0/tests/riak_admin_console_tests.erl.

however, they're not currently part of any group -so they don't get run as part of the release testing.  I will see if they run successfully with this change, and if so add this into the `admin` group of tests so we don't regress this in the future.

martinsumner:
Scratch that previous comment.  The riak_admin_console_tests pass without this change, so clearly they don't detect the problem

martinsumner (approved):
I'm having problems reproducing the original problem, but can confirm this copied correctly from nodetool

PR #1049, opened 2020-12-30
erl scripts

martinsumner:
Some riak admin scripts require an erlang node to be started, and then connected to the riak node (rather than using rpc).

These scripts don't work in 3.0.1.

$ERTS_PATH and $NAME_PARAM are not defined (as was previously defined in node_package);
boot file - cannot find start.boot only start_clean.boot (for reasons unknown);
observer not an installed application

These changes work with OTP 20.3, OTP 21.3 and OTP 22.3.  Tested using:

`make rel`
`rel/riak/bin/riak start`
`rel/riak/bin/riak admin top`


Bob-The-Marauder (approved):
Tested by building a package on CentOS 7. Works fine as far as I can tell. `riak admin top` came up correctly (`riak top` isn't configured but we can live without that). I was pleased to notice that other functions such as `riak admin test` also work now.

PR #1051, opened 2021-01-20
Update riak-admin

martincox:
Pipe json props string to tr to truncate whitespace.


martinsumner (approved):
Tested locally - works for me

PR #1069, opened 2021-05-19
Update riak admin usage and examples to reflect changes

JMercerGit:
Since the format of `riak-admin` was changed to `riak admin`, examples such as the usage prompt from running `riak admin` need to be corrected.
This closes issue #1068

### DIFF
--- a/rel/files/riak-admin
+++ b/rel/files/riak-admin
@@ -9,7 +9,7 @@ ORIGINAL_DIR=$(pwd)
 cd $RUNNER_BASE_DIR
 
 # Identify the script name
-SCRIPT="riak-admin"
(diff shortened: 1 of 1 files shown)


Bob-The-Marauder (commented):
I think you may have been a little over-zealous here. e.g. `SCRIPT="riak-admin"` should be correct as is. The rest look good to me though

JMercerGit:
> I think you may have been a little over-zealous here. e.g. `SCRIPT="riak-admin"` should be correct as is. The rest look good to me though

Thanks, My mistake. I've corrected above.

Bob-The-Marauder (commented):
Looks good to me now

PR #1075, opened 2021-06-01
fix: Remove - flag from su in riak startup script

jrogov:
This is more of a request for comments than an actual PR.

Recently while tweaking [our base image](https://github.com/rbkmoney/image-riak-base) we've discovered that it's practically impossible to pass env variables (such as `WAIT_FOR_ERLANG`) from script.

I believe the reasons for this are the `su` lines with env passing turned off explicitly (via `-` aka `-l` flag) here:
- https://github.com/basho/riak/tree/ff65b0a2ac63dcb5fca5add95cb1b72bb7e8676a/rel/files/riak#L42
- https://github.com/basho/riak/tree/ff65b0a2ac63dcb5fca5add95cb1b72bb7e8676a/rel/files/riak#L52

As well as `su`d commands in node_package:
- https://github.com/basho/node_package/tree/a541c79bb65fb4ac6f60197c333f841065d0aff6/priv/base/env.sh#L246

As a workaround, [preliminary `su` to runner user (`riak`) in this case was used with passed env](https://github.com/rbkmoney/image-riak-base/pull/20/files).

It left us wondering: was it done by purpose? Safety net of sorts to ensure safe environment?

### DIFF
--- a/rel/files/riak
+++ b/rel/files/riak
@@ -39,7 +39,7 @@ else
     fi
     case "$1" in
         start|console|foreground)
-            su - riak -c "RELX_CONFIG_PATH=${RELX_CONFIG_PATH} VMARGS_PATH=${VMARGS_PATH} RUNNER_LOG_DIR=${RUNNER_LOG_DIR} PIPE_DIR=${PIPE_DIR} ${COMMAND} ${*} -pa {{platform_lib_dir}}/patches"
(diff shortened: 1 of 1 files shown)


martinsumner:
I'm the `git blame` for the first of those, and honestly I don't know.  I suspect I've copied something across, rather than having a specific purpose.

@martincox, @yorkshire-steve - do you have any opinions on this?

Bob-The-Marauder:
Not sure if it's relevant but on the theme of waiting for Erlang, can you use `riak-admin wait-for-service riak_kv riak@<nodeIP>` as a work around? See https://www.tiot.jp/riak-docs/riak/kv/2.2.6/using/admin/riak-admin/#wait-for-service for details.

martinsumner:
https://github.com/basho/riak/discussions/1088

PR #1077, opened 2021-07-01
Add pre_start to set as interactive

martinsumner:
Returns riak to pre-3.0 default of interactive mode, by setting a pre-start script.

Should be possible to override by setting the environment variable before starting (if one wishes to run embedded).  Required for user-added code (e.g. to find user-defined map/reduce functions)


Bob-The-Marauder (commented):
Looks good to me

PR #1079, opened 2021-08-29
pass NODETOOL_NODE_PREFIX to relx-generated script, spare some atoms

hmmr:
Each time `riak status` (and the like) is run, relx generates a unique id which is then converted to an atom (of the form "maint1a2b3c4d-riak@127.0.0.1") in the riak VM to which it connects, eventually causing atom table exhaustion.  Now that relx has been [modified](https://github.com/erlware/relx/pull/871) to optionally produce a static id if `NODETOOL_NODE_PREFIX` env var is present instead of generating a random id, let's make use of this. Convenient when the caller guarantees calls are always serialized.

### DIFF
--- a/rel/files/riak
+++ b/rel/files/riak
@@ -39,17 +39,17 @@ else
     fi
     case "$1" in
         start|console|foreground)
-            su - riak -c "RELX_CONFIG_PATH=${RELX_CONFIG_PATH} VMARGS_PATH=${VMARGS_PATH} RUNNER_LOG_DIR=${RUNNER_LOG_DIR} PIPE_DIR=${PIPE_DIR} ${COMMAND} ${*} -pa {{platform_lib_dir}}/patches"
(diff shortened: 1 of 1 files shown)


ThomasArts:
Could you please explain here what happens if you do not set NODETOOL_NODE_PREFIX today and what would happen with your patch if you do not set this variable?

Would it impact customers that don't bother using this env variable?

hmmr:
If `NODETOOL_NODE_PREFIX` is not set, relx will default to [old behaviour](https://github.com/erlware/relx/pull/871/files#diff-83de12d664d5b857812b510f573d00a037755398f06ff4633a72fbece76e6a7fR376), which is to generate a random id. 

Customers are only to be advised to use this feature if they can guarantee their calls to `riak status` are always separated in time. However, if they _can_ guarantee that, the patch will help to mitigate against atom table overflow.

hmmr:
Just a heads-up (apologies if you have already taken this into account): this feature should be mentioned in release announcement. Those who do periodic, and frequent enough, `riak status` on their clusters, would appreciate it.

martinsumner:
@hmmr - release PR is here https://github.com/basho/riak/pull/1082, including release notes update

PR #1097, opened 2021-12-21
WIP Upgrade to rebar3 3.18, with improvements in packaging and release assembly

hmmr:
This is a patchset inspired by the work I have been doing for some months on [Riak CS](https://github.com/TI-Tokyo/riak_cs/commits/develop-3.0)/[Stanchion](https://github.com/TI-Tokyo/stanchion/commits/develop-3.0) 3.0. It includes:

* Upgrade to rebar3 3.17 + a [patch fixing a call to `git revlist`](https://github.com/erlang/rebar3/pull/2654) in repos with same-name branch and tag, and eventually, to 3.18 (where that patch has been merged). The bug this patch addresses is triggered when (for example) a build is attempted on dockerhub.com.
* Brush up packaging for rpm- and deb-based distros:
  - all now include riak.service unit files and require systemd;
  - the full list of distros where packaging was verified now includes Centos 7 and 8, Debian 11 and Amazon Linux 2;
  - `make package` now works on OSX 14, where it is essentially the result of `make rel`, tarred.
* Get [rid of the `-` option](https://github.com/TI-Tokyo/riak/commit/debd415d44574582391bf6ee0211c11a441ec024) to `su` in the launcher script, which removes the su barrier for env vars (something requested in [another PR](https://github.com/basho/riak/pull/1075)). 
* Avoid [copying leveldb git history](https://github.com/TI-Tokyo/riak/commit/3db7582acc7cb8f833031f6502d3fe9dc5cd7040) when making rel- targets.
* Refactoring of the release assembly:
  - make `make rel` relocatable again (specifically, to enable OSX package);
  - make `riak-admin command` equivalent to `riak admin command` (also applies to `riak-repl` and `riak-debug`);
  - point [cuttlefish](https://github.com/TI-Tokyo/cuttlefish/commits/develop-3.0) and [rebar3_cuttlefish](https://github.com/TI-Tokyo/rebar3_cuttlefish/commits/hmmr/vernemq-master/prepare-for-riak-cs-3.0) to TI-Tokyo repos, with support for fqdn type and bin_script template synchronized with that in relx-4.5.0.

There are no changes in Erlang code.

Testing was done as follows:
* on each of the OSes covered, Riak package is built natively, with `make package`, and installed;
* Riak CS and Stanchion packages are built and installed, just as above;
* `riak pid`, `riak ping`, `riak-admin status` work as expected;
* `riak-admin test` passes;
* `riak-admin member-status` and other cluster-related commands work;
* `riak-debug` succeeds;
* `riak-cs-admin test` succeeds;
* Riak can be built with [this Dockerfile](https://github.com/TI-Tokyo/riak_cs_service_bundle/blob/develop/docker/Dockerfile-riak-3.x) and run as part of Riak CS suite in [Riak CS Docker Bundle](https://github.com/TI-Tokyo/riak_cs_service_bundle).


rel/files/riak
```diff
@@ -1,57 +1,87 @@
...
+	# we need to cd to a dir containing the erlang cookie
+	# (previously implicitly done by su -, which option we have
+	# removed in order to allow any env vars to be available for
+	# the ultimate invocation of riak/riak-cs/stanchion)
+	cd {{platform_base_dir}}
+	$SU riak -c "$*"
+    fi
+}
 
+case "$1" in
```
martincox, line 47: Why does each command need to be present in this case - is this not already handled in the extended start script as extensions?

hmmr: It's an effort to make hyphenated and non-hyphenated forms equivalent.

martincox: I don't think it works as expected, and I'm not even sure that resurrecting the hyphenated commands is warranted - what purpose does it serve? It's been a couple of years since we moved to start script extensions. You now have to re-insert a bunch of boilerplate environment stuff at the top of the script files to make them work, whereas that was all inherited with from the root start script with extensions.

My opinion is that it's a backwards step, unless there's a compelling reason that I'm not aware of / considering.

The tidy up across the rest of the release config is very decent though, and much needed. Some of the deps currently reference TI forks, which would need to target basho, but guess that would be fixed up before everything gets merged in.

hmmr: To deal with the easier part first: re-pinning deps to TI-Tokyo repos has a single purpose: pull in cuttlefish and rebar3_cuttlefish, not so much for the fqdn data type in cuttlefish but rather for several commits in the plugin that make it compatible with rebar3 3.18 (by ioolkos, see [this one](https://github.com/TI-Tokyo/rebar3_cuttlefish/commit/57ef2c6bdb7b535d9423bfcde70a0057d8a5a73e) and thereabout), [synchronize](https://github.com/basho/rebar3_cuttlefish/commit/31d14b67fe8cc1173f11667873ea1b617db5ea06) its internal install_update_escript with the ["upstream" version](https://github.com/erlware/relx/blob/main/priv/templates/install_upgrade_escript) of it in relx, and last but not least, this [little gem](https://github.com/basho/rebar3_cuttlefish/commit/2c052177b7ba3e18664ae1b309d78342de9f19d1).

I note that the fqdn-type patch has already been [merged](https://github.com/basho/cuttlefish/commit/95a96b35e533e803cdcfc5dfd09dd308697840c9) in basho/cuttlefish, but there is no tag. If you do the tagging, I will then prepare the necessary PRs to bump cuttlefish version where necessary (it is an explicit dependency in bitcask, clique, eleveldb and riak_sysmon), and collect a patchset for another PR in rebar3_cuttlefish.

Regarding the phasing out of hyphenated forms, I have no strong opinions and will gladly follow the party line (although I would invite @Bob-The-Marauder to this discussion, who has been in the business of dealing with customers for some time).

For the record, my understanding of the underlying machinery is as follows:

* The `riak` script generated by relx and/or cuttlefish provides basic commands like start, stop, console, pid, ping, rpc. It is how riak is started/stopped in riak_test setting. It is not under our control.
* The other `riak`, the one created from the template in rel/files, has a distinct and specific purpose:
  - it takes care of the pid file (relx-generated one will not do this for us);
  - it does `su` as appropriate, and is therefore installed in PATH when riak is installed from a package;
  - it launches `riak-admin` etc when called as `riak admin`.

I saw this last item precisely as a way to enable non-hyphenated forms (if we remove riak-admin and its hyphenated friends from PATH, the move to hide hyphenated forms will be complete, from the user standpoint). Apparently, by "move to start script extensions" you mean to do it differently -- perhaps by injecting code in the script produced by relx? If so, I would be happy to rework my contribution accordingly, but I didn't see how relx can be massaged to achieve that.

martinsumner:
This seems to be failing for me to do both `make rel` and `make devrel`.

e.g. 

```
make rel
rel/riak/bin/riak start
rel/riak/bin/riak ping
```

Returns no pong.  Starting with `riak console ` is OK.

hmmr:
@martinsumner Indeed, riak doesn't start on the very first run (it somehow mends itself the second time and starts just fine on subsequent attempts though). Something to investigate -- thanks for alerting me to this.

In the meantime, can you add a tag (2.1.1 or maybe 2.2.0) somewhere in basho/cuttlefish post https://github.com/basho/cuttlefish/pull/249? This will help to get rid of non-basho dependencies in this PR.

martinsumner:
Could you use the branch `develop-3.0` reference on basho/cuttlefish in rebar.config?  As a general rule, we try and use branch names on dependencies until the point we're cutting releases, then we go through and tag everything in one sweep at that point.

You're right that riak post `make rel` does work on the second start.

From a timeline perspective,  I'm pulling together release 3.0.10 at the moment, just doing some specific performance tests required for the release.  If we can solve any outstanding issues soon, I will check to see if we can add this PR in.  The issue is finding someone to review/approve, as I don't feel confident I understand the changes well enough to OK it myself.

martinsumner:
Some other issues:

- The `riak admin` non-hyphenated flavour doesn't appear to work.  I though the intention was to make both hyphenated and non-hyphenated forms work.
- I'm also getting errors with the hyphenated form with `make rel` releases e.g. `rel/riak/bin/riak-admin cluster status` returns `rel/riak/bin/riak-admin: line 289: relx_nodetool: command not found`

Bob-The-Marauder:
Sorry for being late to the party but regarding the hyphenated/non-hyphenated command issue, having both options working in at least 3.0 is strongly beneficiary for upgrading users who use automated scripts. If the hyphenated commands work then their scripts just continue working as before and there is nothing for them to worry about despite the scariness of a v2 to v3 upgrade. As Martin Sumner will probably verify, one of the common 3.0.x issues found in the Post Riak Slack channel is people trying to use the hyphenated commands in KV 3.0 and getting unexpected errors.

Despite the fact we are about to hit 3.0.10 (at time of writing), there are a lot of people currently on early 2.9.x releases who are currently considering upgrading to 3.0.x within the next 6 months to 18 months. Giving them the smoothest possible transition is probably the best way to gain continued loyalty to the platform.

If there is a really strong need to move exclusively to the non-hyphenated commands, I suggest we put a note in the version 3.0 documentation (currently being written) saying that the hyphenated commands will be deprecated and that from version 3.x.0 e.g. 3.1.0 or 3.2.0 that they will no longer be supported.

martincox:
It's not that there's a strong need to remove the hyphenated scripts. More of a case that they have already been removed and the changes absorbed by two orgs with the largest deployments of riak.

Adding them back adds a bunch of extra boilerplate to scripts and counters the usefulness of start script extensions - there's no point in maintaining both together. 

Perhaps if we had honoured both when the change was made, and eventually deprecated hyphenated versions, it might have been more preferred. But re-introducing several releases later makes little sense to me, and I'd strongly recommend we don't.

hmmr:
To clarify:

* When installed from a package, hyphenated and non-hyphenated forms are equivalent. In this case, the `riak` executable that gets installed in /usr/sbin comes from the template in rel/files/riak. It takes care of pid files and ensures the pipe dir exists, routes `riak admin` to `riak-admin` etc, drops privileges to riak user, and otherwise forwards options to /usr/lib/riak/bin/riak (the script generated by relx). All hyphenated forms are installed in /usr/sbin.
* When built locally with `make rel`, nothing gets installed outside rel/riak. The riak executable is generated by relx (and _not_ by rebar, from rel/files/riak), and that is copied into rel/riak/bin, alongside riak-admin, riak-debug, riak-chkconfig and riak-repl (duly generated from templates in rel/files). Here, only hyphenated forms of `riak-admin` etc exist.
* The "build locally" case would also cover a docker-based build.

I personally have no strong opinions here; it does seem to me, however, that maintaining hyphenation equivalence is good for 'pure' customers (those who install riak from packages and don't build riak themselves), and we should continue to offer the equivalent forms -- especially given that it's already there and comes at no cost.

martincox:
If it is absolutely necessary to bridge the gap of hyphenated forms being deprecated, why not take a less invasive route and create proxy scripts for the hyphenated forms that live in /usr/sbin/ with something in the form of:

```
#!/bin/sh
echo "riak-admin command is now deprecated, use riak admin bla bla"
riak admin $@
```

It saves trying to shoehorn the hyphenated forms back into the release config, which will ultimately be removed later anyway. As I mentioned before, I see little point in maintaining both. An even simpler workaround might be to use aliases.

With regards to the PID writing - does that not work? It should be handled as part of the start script. I mean, I've just checked over a few clusters (3.*) and all have PIDs in the expected directory.

martincox:
Also, just for my own clarity, what branch are we targeting with this PR? It's set to develop-3.0, but I'm not sure if is because that happened to be the default branch at the time it was created.

hmmr:
> This seems to be failing for me to do both `make rel` and `make devrel`.
> 
> e.g.
> 
> ```
> make rel
> rel/riak/bin/riak start
> rel/riak/bin/riak ping
> ```
> 
> Returns no pong. Starting with `riak console ` is OK.

Fixed in https://github.com/basho/riak/pull/1097/commits/e6f64bfadf128d86c14a7ee1b68d4e0adb42d087

hmmr:
@martincox Re:

> If it is absolutely necessary to bridge the gap of hyphenated forms being deprecated, why not take a less invasive route and create proxy scripts for the hyphenated forms that live in /usr/sbin/ with something in the form of:

With https://github.com/basho/riak/pull/1097/commits/1ce026214e75470afb87f47929b45b5bd05ab51b, when installed from a package on a debian11 system, we now have:

```
root@debian:~# riak admin
Usage: riak-admin { cluster | join | leave | backup | restore | test | 
                    reip | erl-reload | wait-for-service | 
...
root@debian:~# riak-admin
The hyphenated form of this command is deprecated. Please fix your scripts to call `riak admin` instead.

Usage: riak-admin { cluster | join | leave | backup | restore | test | 
                    reip | erl-reload | wait-for-service | 
                    ringready | transfers | force-remove | down |
```

> With regards to the PID writing - does that not work? It should be handled as part of the start script. I mean, I've just checked over a few clusters (3.*) and all have PIDs in the expected directory.

It's another possible regression. I have the `extended_start_script_hooks` stanza in relx/deb section in rebar.config, unchanged since 3.0.9, but if I delete [this line](https://github.com/TI-Tokyo/riak/blob/hmmr/develop-3.0/upgrade-to-rebar3.17/rel/files/riak#L57), the pid file is not written. `riak pid` still works (pid is obtained by calling `os:getpid/0`, so no wonder). I will investigate.

martinsumner:
@hmmr wanted to try test this today, but had issues fetching the referred to tag in riak_repl.  Is the TI-Tokyo riak_repl repo public?

hmmr:
@martinsumner It is indeed private. The change I pushed in TI-Tokyo repo is just `s|git://|https://|` on top of tag `riak_kv-3.0.9`. I guess you can do this change manually in the official repo, tag it `riak_kv-3.0.10`, and I will then re-point riak_repl to the new tag accordingly in this PR.

I have traced the pid file regression, which was in rebar3_cuttlefish plugin (it was introduced by vernemq people in their more recent changes, and adopted by too eager me). I have reverted to version 0.2.0, tweaking it to work with rebar3.18 (and noticed that @martincox has already [done](https://github.com/basho/rebar3_cuttlefish/commit/177a06ea1bec973864365470f14a34820b95dfc6) the same). I have just re-pointed rebar3_cuttlefish to basho/develop branch, and it is great again.

There are still issues with https://github.com/basho/riak/pull/1097/commits/f75df62e2e042f2313a387f1580c31eb3bd4b270, which I am trying to fix, but you should be able to build it locally with `make rel`.

hmmr:
Superseded by https://github.com/basho/riak/pull/1108.

PR #1122, opened 2022-11-22
Temp replace reip with reip_manual

martinsumner:
Reip was broken by the missing scripts necessary to generate the config so that erl could be run without the riak node active, but with the config necessary to load riak_core.

Now loading riak_core is not necessary with riak_manual .. but both the path to the ring directory, and the cluster name need to be manually passed in as command params.


hmmr:
Works as expected. One minor point: after a successful execution, the node, now with a new node name, needs to have its riak.conf modified to reflect that (otherwise the node will crash at startup). I believe a reminder to change nodename in riak.conf would not hurt, around here: https://github.com/basho/riak_kv/pull/1838/files#diff-97040b499de3444a6d305784c280ce6c35fcb352603d2fb61e9a43957e135380R256.

hmmr:
+1

